PR #16, opened 2021-09-16
feat: track events using SDK

levibostian:
1 event is tracked. Generate a profile using the random generate button. If you do, an event will be added to your profile. If you do not, you will not see the event. ﻿


Remote Habits/Repository/ProfileRepository.swift
```diff
@@ -90,6 +90,6 @@ class AppProfileRepository: ProfileRepository {
             return onComplete(.success(()))
         }
 
-        messagingPush.registerDeviceToken(existingApnDeviceToken, onComplete: onComplete)
+        messagingPush.registerDeviceToken(String(apnDeviceToken: existingApnDeviceToken), onComplete: onComplete)
```
levibostian, line 93: We can consider creating a 2nd `MessagingPush.registerDeviceToken()` function that takes in `Data` as the parameter instead of `String`. Since we are using `registerDeviceToken()` as a public facing function for customers that don't want to use `application(didRegisterRemoteNotifications))` or have a more advanced use case, providing this 2nd function might provide less bugs instead of a customer trying to incorrectly convert from `Data` to `String` when passing APN tokens to `registerDeviceToken`.

hownowstephen: good idea :D I can add that

hownowstephen (approved):

Remote Habits/Repository/ProfileRepository.swift
hownowstephen: out of curiousity, why are you using `MessagingPush` and not `MessagingPushAPN` here?

levibostian: > out of curiousity, why are you using MessagingPush and not MessagingPushAPN here?

Not on purpose. I have not written automated tests against the `ProfileRepository` yet so I have not yet run into the "Oh, wait. I can't mock using this class" scenario. So, I probably *should* be using it.

PR #19, opened 2021-09-24
feat: register FCM device token

levibostian:
Decided to make 2 separate `AppDelegate` to avoid customers thinking they should use both FCM and APN in their app. I have seen this in help tickets in the past and want to help avoid them. 

As RH grows, we may need to eventually create separate example apps. One for APN, one for FCM, one using automated tests, etc. Clear examples for customers to view?


Remote Habits/RemoteHabitsApp.swift
```diff
@@ -2,7 +2,13 @@ import SwiftUI
 
 @main
 struct RemoteHabitsApp: App {
+    /**
+     Here, decide what `AppDelegate` that you would like to use for the app.
+     1. AppDelegate - use for APN push service
+     2. AppDelegateFCM - use for FCM push service
+     */
     @UIApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
```
hownowstephen, line 10: any downside to calling the other one `AppDelegateAPN`?

hownowstephen (approved):

Issue #30, opened 2021-11-29
Create new Remote Habits app for iOS


ami-aman:
Reference mockup designs
https://www.figma.com/file/BJdEVphP1xPOPFQcKkqNhk/Remote-Habits-App?node-id=137%3A304

lj-cio:
Notes on Functionality:

* On Login, identify customer with the `name` as an attribute using the SDK (https://customer.io/docs/sdk/ios/identify/#identify).
* On Logout, stop identifying the customer using the SDK (https://customer.io/docs/sdk/ios/identify/#clearIdentify).
* On click of a "Habit", send event using the SDK with properties `name: Clicked Habit` and property `habit: <habit name>` (https://customer.io/docs/sdk/ios/track-events/#track-a-custom-event).
* On change of the toggle, number of times, time start, or time end for a "Habit", send event using the SDK with properties `name: Habit Updated` and properties `habit: <habit name>`, `frequency: <number of times>`, `start_time: <start time>`, `end_time: <end time>`.

hownowstephen:
when changing site_id/api_key we can use the `dexterity-check` [endpoint](https://github.com/customerio/edge/blob/426634059656115ddcc4d702701a91b5b65b36f7/misc/router.go#L10) at the edge to validate the supplied creds, which is just doing a basic auth GET to https://track.customer.io/dexterity-check 

```
curl -XGET https://track.customer.io/dexterity-check -u siteid:apikey
```

PR #31, opened 2021-12-02
feat: new UI - login and habits selection

ami-aman:
[Create new Remote Habits app for iOS](https://github.com/customerio/RemoteHabits-iOS/issues/30)

**This PR contains :**

Creating new remote habits application for demoing to the customers how to integrate and use CIO SDK.

Current PR has the following :

- UI implementation using UIKit ✅ 
- New designs as per [Prototype](https://www.figma.com/file/BJdEVphP1xPOPFQcKkqNhk/Remote-Habits-App?node-id=137%3A304) ✅
- Show data using Stub data only ✅
- Same Bundle Identifier for new app ✅
- Customer.io SDK integration ✅
- Documentation ❌ 
- Test cases ❌ 
- API Integration ❌

There is scope of improvement which was felt while merging the new app with current RH app. Just to name a few - merging existing code with new code & remove redundant code, re-arrange files & put them at proper place but none of these affect the performance & functioning of the app. 

~~New RH app also makes use of CocoaPods for smooth Floating Labels.~~
~~Please use `pod install` before running the app~~
**REMOVED THE SUPPORT OF COCOAPODS, LIBRARY HAS BEEN INTEGRATED USING SWIFT PACKAGE MANAGER**
.
.
Simulator Screen Shot - iPhone 13 Pro Max - 2021-12-02 at 18 28 18Simulator Screen Shot - iPhone 13 Pro Max - 2021-12-02 at 18 28 21Simulator Screen Shot - iPhone 13 Pro Max - 2021-12-02 at 18 28 42


hownowstephen (commented):
Looking great to me, @ami-aman! Will leave the approval on this one to @levibostian in case he has any specific feedback, since I'm still a Swift novice - seems like the right approach to split out the SDK integration piece into follow-up work :smile:  looking forward to being able to use this one for testing

ami-aman:
@hownowstephen  thanks for the quick review. I have merged the code in such a way that it will not need any further integration of SDK. Starting from tomorrow my time, I will directly start integrating APIs and will make small commits/PRs haha and not giant ones like this.

RussellCIO:
`sudo gem install cocoapods` to install CocoaPods if you're not using a RVM. Otherwise, you can find more info here for alternative installations for anyone testing this pr. https://guides.cocoapods.org/using/getting-started.html

levibostian:
I see `.DS_Store` files committed. Make sure to please remove those.

Remote Habits/View/Common/RHBaseViewController.swift
```diff
@@ -0,0 +1,90 @@
...
+    }
+    
+    @objc func goBack() {
+        self.dismiss(animated: true, completion: nil)
+    }
+
+    func configureNavigationBar(titleColor: UIColor = .black, backgoundColor: UIColor = .white, title: String, hideBack : Bool, showLogo : Bool = false) {
+        if hideBack {
+            self.navigationItem.setHidesBackButton(hideBack, animated: true)
+        }
```
levibostian, line 37:
```suggestion
        self.navigationItem.setHidesBackButton(hideBack, animated: true)
```

Since you are passing in `hideBack` Bool value anyway, no need for the `if {}`. Also, I imagine a bug could happen if you call `configureNavigationBar` with `hideBack = true` and then call it again with `hideBack = false`.

Remote Habits/View/Extension + Customisation/RHCustomisedTableView.swift
```diff
@@ -0,0 +1,28 @@
...
+//  RHCustomisedTableView.swift
+//  Remote Habits Mobile App
+//
+//  Created by Amandeep Kaur on 01/12/21.
+//
+
+import Foundation
+import UIKit
+
+class RHCustomisedTableView : UITableView {
```
levibostian, line 11: What is this file for? Some code comments would be helpful for future reference. I don't understand the need of this file.

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+import UIKit
+
+class RHDashboardViewController: RHBaseViewController {
+
+    // MARK: - --OUTLETS--
+    @IBOutlet weak var dashboardTableView: UITableView!
+    
+    // MARK: - --VARIABLES--
+    var dashboardData  = UserHabit()
+    var isLoggedIn : Bool = true // Change this by checking logged in user detail using CIO SDK
```
levibostian, line 17: Is this variable just a stub for now? I don't see it modified anywhere in this file?

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+        // Do any additional setup after loading the view.
+    }
+    
+    // MARK: - --FUNCTIONS--
+    func setupDashboardTableView() {
+        
+        dashboardData = RHStubData().getStubData(isLoggedIn: isLoggedIn)
+        
+        // HabitTableViewCell move to constant file
+        dashboardTableView.register(UINib(nibName: RHConstants.kHabitTableViewCell, bundle: nil), forCellReuseIdentifier: RHConstants.kHabitTableViewCell)
```
levibostian, line 37: Could we turn this into an extension? It's boilerplate code to me. 

Something like this? 
```swift
dashboardTableView.register(nib: RHConstants.kHabitTableViewCell)
```

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+        dashboardTableView.rowHeight = UITableView.automaticDimension
+        dashboardTableView.estimatedRowHeight = 80
```
levibostian, line 39: When setting auto row height, I always forget something. Maybe an extension would be good here, too? 

Something like this? 
```swift
dashboardTableView.setAutorowHeight()

// extension function
extension UITableView {
  func setAutorowHeight(estimatedRowHeight: Int = 40) {
    ...
  }
```

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+            label.frame = CGRect.init(x: 16, y: 60, width: headerView.frame.width-10, height: 26)
+        }
+        
+        label.text = headerData?.headerTitle
+        label.font = UIFont(name: headerData?.titleFontName ?? RHConstants.kEmptyValue, size: CGFloat(headerData?.titleFontSize ?? 17))
+        label.textColor = .black
+        
+        headerView.addSubview(label)
+        
+        return headerView
```
levibostian, line 76: Could this code move into a custom `UIView`? Like the other `UITableViewCell`s in this project are?

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+    }
+    
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        
+        if indexPath.section != 0 { return }
```
levibostian, line 81: I found this code take a bit for me to understand what it's doing. Maybe something like this makes it more readable? 

```suggestion
        let isTableViewCell = indexPath.section == 0
        if !isTableViewCell { return }
```

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+        return cell
+    }
+    
+    func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+        return UITableView.automaticDimension
+    }
+    
+    func tableView(_ tableView: UITableView, estimatedHeightForRowAt indexPath: IndexPath) -> CGFloat {
+        return 80
+    }
```
levibostian, line 137: Are these needed since `dashboardTableView.rowHeight = UITableView.automaticDimension` is called above in this class?

Remote Habits/View/RHHabitDetailViewController.swift
```diff
@@ -0,0 +1,117 @@
...
+    let newHeaderViewHeight: CGFloat =
+              headerViewHeightConstraint.constant - y
+    if newHeaderViewHeight > headerViewMaxHeight {
+       headerViewHeightConstraint.constant = headerViewMaxHeight
+    } else if newHeaderViewHeight <= headerViewMinHeight {
+       headerViewHeightConstraint.constant = headerViewMinHeight
+    } else {
+       headerViewHeightConstraint.constant = newHeaderViewHeight
+       scrollView.contentOffset.y = 0 // block scroll view
+    }
```
levibostian, line 115: What is this code doing? I find it difficult to read. Maybe some function naming and/or extensions would help?

Remote Habits/View/RHLoginViewController.swift
```diff
@@ -0,0 +1,160 @@
...
+    }
+    
+    func setUpTextFields() {
+        for field in textFields {
+            
+            switch(field) {
+            case userNameInput :
+                field.iconText = "\u{f007}"
+            case emailInput:
+                field.iconText = "\u{f0e0}"
```
levibostian, line 63: Should these strings be turned into constants?

Remote Habits/View/RHLoginViewController.swift
```diff
@@ -0,0 +1,160 @@
...
+    // MARK: - --VARIABLES--
+    var textFields: [SkyFloatingLabelTextFieldWithIcon] = []
+    
+    // MARK: - --LIFECYCLE METHODS--
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        // Do any additional setup after loading the view.
+        
+        addLoginBackground()
+        textFields = [userNameInput, emailInput]
```
levibostian, line 30: This seems somewhat hard to maintain in the future. Could we do something like this? 

```swift
func setupTextFields() {
  self.textFields = []
  for view in views {
    if let textField = view as? UITextField {
      textFields.append(textField)
    }
  }
}
```

Remote Habits/View/RHLoginViewController.swift
```diff
@@ -0,0 +1,160 @@
...
+        return name.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines) != RHConstants.kEmptyValue
+    }
+    
+    func validateEmail(_ email: String) -> Bool {
+        
+        let emailRegEx = "[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,64}"
+
+        let emailPred = NSPredicate(format:"SELF MATCHES %@", emailRegEx)
+        return emailPred.evaluate(with: email)
+    }
```
levibostian, line 92: Could we move these functions into a `Util` class or something? Move it out of the `ViewController`?

Remote Habits/View/RHLoginViewController.swift
```diff
@@ -0,0 +1,160 @@
...
+    
+    func route(withParam isGuestLogin : Bool = true) {
```
levibostian, line 94: Not sure why param is named `withParam`? 

```suggestion
    func route(isGuestLogin : Bool = true) {
```

Remote Habits/View/RHLoginViewController.swift
```diff
@@ -0,0 +1,160 @@
...
+        route(withParam: false)
+    }
+    
+}
+
+
+// MARK: - UITextFieldDelegate
+extension RHLoginViewController : UITextFieldDelegate {
+    func textFieldDidEndEditing(_ textField: UITextField) {
+        if textField == emailInput && textField.text?.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines) == RHConstants.kEmptyValue {
```
levibostian, line 154: An extension could be useful here

```swift
textField.textStrippedWhitespace

extension UITextField {
  var textStrippedWhitespace: String {
     textField.text.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines) ?? ""
  }
}
```

Remote Habits/View/RHSwitchWorkspaceViewController.swift
```diff
@@ -0,0 +1,100 @@
...
+    
+    // MARK: - --FUNCTIONS--
+    func setUpMainView() {
+        mainView.layer.cornerRadius = 13
+    }
+    
+    func setUpWorkspaceButton() {
+        switchWorkspaceButton.titleLabel?.font = RHFont.SFProTextSemiBoldMedium
+        switchWorkspaceButton.setTitleColor(UIColor.lightGray, for: .disabled)
+        switchWorkspaceButton.setTitleColor(UIColor.white, for: .normal)
```
levibostian, line 40: Can we move this into a custom `UIButton` class? Make this code re-usable?

Remote Habits/View/RHSwitchWorkspaceViewController.swift
```diff
@@ -0,0 +1,100 @@
...
+                    switchWorkspaceButton.isEnabled = true
+                }else {
+                    switchWorkspaceButton.isEnabled = false
+                }
+            }
+        }
+    }
+    
+    func validateInput(with text : String) -> Bool{
+        return text.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines) != RHConstants.kEmptyValue
```
levibostian, line 78: Can we move this into an extension or util class? Move the logic out of the ViewController?

Remote Habits/View/Xib/HabitDetail/HabitAddInfoTableViewCell.swift
```diff
@@ -0,0 +1,25 @@
...
+
+import UIKit
+
+class HabitAddInfoTableViewCell: UITableViewCell {
+
+    @IBOutlet weak var mainCellView: UIView!
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        // Initialization code
+        mainCellView.layer.cornerRadius = 13
```
levibostian, line 16: I have seen `cornerRadius = 13` called many times in this code. Sounds like 13 is a radius we want to use in lots of views here. 

Perhaps it's best to create some custom View subclasses and/or use constants for this code?

Podfile
```diff
@@ -0,0 +1,27 @@
+# Uncomment the next line to define a global platform for your project
```
levibostian, line 1: Comment here simply as a reminder for switching to SPM over cocoapods for installing SkyFloatingLabelTextField.

Remote Habits/AppDelegate.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  AppDelegate.swift
+//  Remote Habits Mobile App
+//
+//  Created by Amandeep Kaur on 25/11/21.
+//
```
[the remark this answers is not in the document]
ami-aman: This would be done in next PR. Already listed as a TODO item and will be taken care in the next PR raised.

Remote Habits/Common/RHConstants.swift
```diff
@@ -0,0 +1,37 @@
+//
+//  RHConstants.swift
+//  Remote Habits Mobile App
+//
+//  Created by Amandeep Kaur on 25/11/21.
+//
+
+import Foundation
+
+struct RHConstants {
```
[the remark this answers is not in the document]
ami-aman: RH stands for Remote Habits. While I was working on a separate project (before merging) I was following this approach and I personally like this because it gives a sense of belongingness of files in the app. I was expecting a replacement of current repo with new repo itself but that was ruled out in our discussions. So to keep the work going I am still using RH, but as mentioned in the PR description, I see scope of improvement while merging and this is one of them.

Remote Habits/Common/RHConstants.swift
```diff
@@ -0,0 +1,37 @@
...
+    // Strings
+    static let kBackground = "background"
+    static let kInvalidEmail = "Invalid Email"
+    static let kEmptyValue = ""
```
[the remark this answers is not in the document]
ami-aman: Using "" at many places in project. To keep the code look clean using it from constants value instead of writing down "" everywhere. I do understand that functionally "" or .kEmptyValue doesn't make any difference but I like to keep the code clean by using all string values from a constant file.

Remote Habits/Common/RHConstants.swift
```diff
@@ -0,0 +1,37 @@
...
+//  Remote Habits Mobile App
+//
+//  Created by Amandeep Kaur on 25/11/21.
+//
+
+import Foundation
+
+struct RHConstants {
+    // Strings
+    static let kBackground = "background"
```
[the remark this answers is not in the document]
ami-aman: Makes sense. I will update it in the next commit.

Remote Habits/View/RHLoginViewController.swift
ami-aman: This function has been revamped. Will be updating with next PR.

Remote Habits/View/RHLoginViewController.swift
ami-aman: Noted. Will be updated with next PR.

Remote Habits/View/Xib/HabitDetail/HabitAddInfoTableViewCell.swift
ami-aman: Can use constants for this radius. Yes, it is being used in the dashboard and dashboard detail screen.

Podfile
ami-aman: Thank you for the reminder.

Remote Habits/View/RHSwitchWorkspaceViewController.swift
ami-aman: We can but not prioritising it right now since our aim is to demo the app to customer. Regardless this will done by EONextWeek (max).

ami-aman:
I see 5 conflicts now and we are not going to use these files in the new RH app. @levibostian can you confirm if any of those changes are needed by you because in new RH app I am not using two different AppDelegates and ContentView. I can replace these files if you confirm.

Podfile
ami-aman: Removed cocoapods and installed library using SPM. Changes can be found in latest commit

Remote Habits/Common/RHFont+Color.swift
```diff
@@ -0,0 +1,26 @@
...
+    static let SFProTextSemiBoldMedium = UIFont(name: "SFProText-Semibold", size: 17)
+    static let AwesomeFontLarge =  UIFont(name: "FontAwesome", size: 20)
+}
+
+struct RHColor {
+    static let LabelGray = UIColor(red: 0.235, green: 0.235, blue: 0.263, alpha: 0.6)
+    static let LineGray = UIColor(red: 0.776, green: 0.776, blue: 0.784, alpha: 1)
+    static let SelectedLineGray = UIColor(red: 0.650, green: 0.650, blue: 0.650, alpha: 1)
+    static let DefaultBackground = UIColor(red: 247/255, green: 248/255, blue: 250/255, alpha: 1)
+}
```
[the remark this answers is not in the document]
ami-aman: Yes, we can do that. Good suggestion ! But noting it down for now and will make changes in the upcoming PRs.

Remote Habits/Common/RHConstants.swift
```diff
@@ -0,0 +1,37 @@
...
+    
+    // Storyboard / Xib
+    static let kStoryboardMain = "Main"
+    static let kDashboardViewController = "RHDashboardViewController"
+    static let kHabitTableViewCell = "HabitTableViewCell"
+    static let kHabitDetailViewController = "RHHabitDetailViewController"
+    static let kSwitchWorkspaceViewController = "RHSwitchWorkspaceViewController"
+    static let kHabitReminderTableViewCell = "HabitReminderTableViewCell"
+    static let kHabitDetailToggleTableViewCell = "HabitDetailToggleTableViewCell"
+    static let kHabitAddInfoTableViewCell = "HabitAddInfoTableViewCell"
```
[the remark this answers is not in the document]
ami-aman: Same xib is possibly being used in multiple controllers so to keep the code clean, instead of creating a new variable and saving `identifier` in all those files, saving it in constants file and using it.

Pros - In future if there is any need to change value of identifier then it would be one line change, instead of locating and replacing the values in all files.

Remote Habits/Info.plist
```diff
@@ -31,17 +48,27 @@
 	</array>
 	<key>CFBundleVersion</key>
 	<string>39f9d999</string>
+	<key>FirebaseAppDelegateProxyEnabled</key>
```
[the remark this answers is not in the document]
ami-aman: Thanks for pointing it out, changes can be seen in next commit.

Remote Habits/View/Common/RHBaseViewController.swift
ami-aman: There is an else statement to this `if` condition. In case `hideBack` is `false` then the function will display a back button otherwise hide it. Also, if `hideBack` is false then I didn't want any extra code to be done like updating the value of `setHidesBackButton` so it;s written this way.

Remote Habits/View/Extension + Customisation/RHCustomisedTableView.swift
ami-aman: This file is used to add animation on HabitDetailViewController screen. It is used as custom class for tableView on this screen. 

FYI - Animation to expand and collapse header when table is scrolled

Remote Habits/View/RHDashboardViewController.swift
ami-aman: This variable will be removed in future PRs. This PR only has UI functionality so to make login work, this variable was added. Since now API integration is complete so we do not need this property anymore.

Remote Habits/View/RHDashboardViewController.swift
ami-aman: Noted. Will add it in future PRs.

Remote Habits/View/RHDashboardViewController.swift
ami-aman: Yes, we can add an extension for both the properties since it is being used at multiple places. Noting it down to make the changes in future PRs.

Remote Habits/View/RHDashboardViewController.swift
ami-aman: Yes, I have already moved the code from `cellForRowAt indexPath` to declutter Dashboard and will be moving this and other methods too, if needed. Changes will be seen in the next PR.

Remote Habits/View/RHDashboardViewController.swift
ami-aman: Oops. My bad ! Thanks :) Fixed and changes can be seen in this PR last commit

Remote Habits/View/RHHabitDetailViewController.swift
ami-aman: This is already an extension and using the delegate method of ScrollView to expand and collapse the height of header when the bottom tableview is scrolled up and down. But if you find it not so readable I can move the code in another function being called from this delegate method to make it clear for it's functioning. Btw.. i am yet to do comments so once I add them then you can expect it to be better and clear to understand.

Remote Habits/View/RHLoginViewController.swift
ami-aman: Ah ! I didn't make them constants since using it only at one place. Does it make any difference ? @levibostian

Remote Habits/View/RHLoginViewController.swift
ami-aman: I understand your perspective, but I don't think we are going to have any other text field on LOGIN screen ever except these two i.e. username and email id. OR we can have ??? @hownowstephen :thinking_face:

Remote Habits/View/RHLoginViewController.swift
ami-aman: Yep. Make sense. Noting it down for future PRs.

ami-aman:
**P.S**. I am mentioning `in the upcoming PRs` because I  do not want to over clutter this PR as it already has so many file changes. But I have made minor changes to this one too. Going forward I will be raising one functionality, one PR to make review easy.

Also, I want to get this one approved asap so that the product team and QA team can start looking at it and check on it's functionality. If there are suggestions, then I will be able to implement them sooner as I would already be shipping the next PRs. This PR only has UI functionality so if there is something not expected or expected then I will be able be able to handle them with upcoming PRs.

Just that I want to expedite the process and this feel there will be good round of testings before final delivery + we will be able to ship new RH quicker.

Hope that's okay with the team @levibostian @hownowstephen @RussellCIO

hownowstephen (commented):
LGTM, but leaving approval to @levibostian since there's a couple convos for him to :eye: 

re: splitting out functionality, sounds great - helps us to review the additional changes in isolation :+1:

Remote Habits/Base.lproj/LaunchScreen.storyboard
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
levibostian: I agree. I am ok with using storyboards.

Remote Habits/View/Common/RHBaseViewController.swift
levibostian: Ok, I understand that. 

I do believe you will still need to include `self.navigationItem.setHidesBackButton(false, animated: true)` in the `else` condition though, yeah?

Remote Habits/View/Extension + Customisation/RHCustomisedTableView.swift
levibostian: If it's used to add an expended/collapse animation, perhaps it would be a good idea to rename the class to reflect that purpose?

Remote Habits/View/RHHabitDetailViewController.swift
levibostian: I'll let you decide the technique you find best to make the code more readable. It's difficult to read and understand at glance.

Remote Habits/View/RHLoginViewController.swift
levibostian: Sorry for the confusion. I am recommending putting `setupTextFields()` in `RHBaseViewController` so that *all* ViewController classes can use the function. Make *all* ViewControllers in the app then easier to maintain the list of TextFields.

Remote Habits/View/RHLoginViewController.swift
levibostian: I understand your thinking there. 

Constants in this scenario would help label the strings since they are not human readable. maybe something like this? 

```swift
field.iconText = Icons.envelope

enum Icons: String {
  case envelope = "\u{f0e0}"
}
```

levibostian (approved):
LGTM!

I replied to your comments and besides that, we can delete the `Remote Habits.xcworkspace` directory since we are not using cocoapods, yeah? 

I *think* that the workspace was created by the cocoapods step. What happens when you run [these CLI commands](https://stackoverflow.com/a/34765245) to uninstall cocoapods from the project? Does it remove the workspace directory then?

PR #35, opened 2021-12-13
feat: track habit on dashboard, habit detail, change workspace 

ami-aman:
[Create new Remote Habits app for iOS](https://github.com/customerio/RemoteHabits-iOS/issues/30)

**This PR contains :**

- Remaining actions for new RH 

Note - @levibostian and @hownowstephen Number of files to be reviewed is around 16.

**Current PR has the following (Detailed description) :**

- Track Habits from Dashboard :white_check_mark: 
- Track Habit from Detail screen :white_check_mark: 
- Validate and Change Workspace :white_check_mark: 
- Push Notifications and Rich Push Notifications  :white_check_mark:  :white_check_mark: 
- Remote Habits App icon  :white_check_mark: IMG_1020
- Fixes from previous PR comments:- 
-- Removed isLoggedIn property which was no longer used :white_check_mark: 
-- Changed colors to color assets :white_check_mark: 
- Documentation ❌ 
- Test cases ❌ 
- API Integration  :white_check_mark: 

**Testing**

- [x] Manual Testing on Simulator
- [x] Testing on Device

**REMOVED THE SUPPORT OF COCOAPODS, LIBRARY HAS BEEN INTEGRATED USING SWIFT PACKAGE MANAGER**
.
.

Simulator Screen Shot - iPhone 13 Pro Max - 2021-12-10 at 21 38 24      Simulator Screen Shot - iPhone 13 Pro Max - 2021-12-10 at 21 38 28  IMG_1019
IMG_1020


Remote Habits/Model/DashboardData.swift
```diff
@@ -24,13 +24,17 @@ struct HabitData {
...
-    let isHabitEnabled : Bool?
-    let actionButtonValue : String?
-    let actionType : HabitActionType?
+    var isHabitEnabled : Bool?
+    var frequency : Int?
+    var startTime : String?
+    var endTime : String?
+    var description : String?
+    var actionButtonValue : String?
+    var actionType : HabitActionType?
```
levibostian, line 37: Is there a reason the struct is changing to `var` instead of `let`? I prefer making structs immutable. I suggest if you need a struct to be modified, use:

```swift
struct HabitDetail: AutoLenses {
```

And sourcery will generate a design pattern that allows you to more safely modify immutable structs. 

You will be able to use the generated code like:
```swift
var habit: HabitDetail = ...
habit = habit.frequenceySet(nil).descriptionSet("")
```

Remote Habits/Model/Profile.swift
```diff
@@ -3,3 +3,12 @@ import Foundation
 struct Profile {
     let email: EmailAddress
 }
+
+struct ValidateWorkspaceResponse: Codable {
+    let meta: CredMessage?
+}
+
+struct CredMessage : Codable {
```
levibostian, line 11: What does `Cred` mean?

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -98,42 +115,56 @@ extension RHDashboardViewController : UITableViewDataSource {
...
         return cell
     }
+    
+    func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+        return UITableView.automaticDimension
+    }
+    
+    func tableView(_ tableView: UITableView, estimatedHeightForRowAt indexPath: IndexPath) -> CGFloat {
+        return 80
+    }
```
levibostian, line 137: I think I remember you changing to using an extension in a previous PR, right? No more need for this code?

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -98,42 +115,56 @@ extension RHDashboardViewController : UITableViewDataSource {
...
+                    break
+                }
+            }
+            
+            if isPushReqd {
+                let main = UIStoryboard(name: RHConstants.kStoryboardMain, bundle: nil)
+                let vc = main.instantiateViewController(withIdentifier: RHConstants.kLoginViewController) as! RHLoginViewController
+                navigationController?.setViewControllers([vc], animated: true)
+            }
+        }
```
levibostian, line 167: Why is this code needed? It seems a bit complex for simply showing a ViewController?

Remote Habits/View/RHHabitDetailViewController.swift
```diff
@@ -42,6 +47,16 @@ class RHHabitDetailViewController: RHBaseViewController {
         habitDetailTableView.setContentOffset(.zero , animated: false )
 
     }
+    
+    
+    func setUpHeader() {
+        reminderLabel.text = "Set reminder to \(habitDetailData?.title ?? "good remote habits")"
```
levibostian, line 53: Since we are building this app from scratch and it's early on, is now a good time to use localized strings over hard coded English ones? I figure now might be the best time to implement this best practice while the app is small? 

At least we should create a ticket in the tech debt backlog to do this in the future?

Remote Habits/View/RHHabitDetailViewController.swift
```diff
@@ -66,23 +81,40 @@ extension RHHabitDetailViewController : UITableViewDataSource {
             guard let cell = tableView.dequeueReusableCell(withIdentifier: RHConstants.kHabitDetailToggleTableViewCell, for: indexPath) as? HabitDetailToggleTableViewCell else {
                 return UITableViewCell()
             }
-
+            cell.actionHandler = self
+            cell.selectionStyle = .none
```
levibostian, line 85: Is there a scenario where `selectionStyle` will *not* be `.none`? If not, perhaps we should move that code into `HabitDetailToggleTableViewCell`. 

Same goes for the rest of the cells below.

Remote Habits/View/RHHabitDetailViewController.swift
```diff
@@ -66,23 +81,40 @@ extension RHHabitDetailViewController : UITableViewDataSource {
...
         return UITableViewCell()
     }
+    
+    func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+        return UITableView.automaticDimension
+    }
+    
+    func tableView(_ tableView: UITableView, estimatedHeightForRowAt indexPath: IndexPath) -> CGFloat {
+        return 100
+    }
```
levibostian, line 117: I think I remember you changing to using an extension in a previous PR, right? No more need for this code?

Remote Habits/View/RHHabitDetailViewController.swift
```diff
@@ -107,3 +139,27 @@ func scrollViewDidScroll(_ scrollView: UIScrollView) {
...
+    func toggleHabit(toValue isEnabled: Bool) {
+        
+        let activity = isEnabled ? RHConstants.kHabitEnabled : RHConstants.kHabitDisabled
+        let selectedHabitActivity = SelectedHabitData(title: habitDetailData?.title,
+                                                      frequency: habitDetailData?.habitDetail?.frequency,
+                                                      startTime: habitDetailData?.habitDetail?.startTime,
+                                                      endTime: habitDetailData?.habitDetail?.endTime)
+        
+        trackerViewModel.trackHabitActivity(withName: activity, forHabit: selectedHabitActivity)
+    }
```
levibostian, line 154: This code seems similar to [this ViewController code](https://github.com/customerio/RemoteHabits-iOS/blob/e5bb87f5afb1d324e91c023f047078dde93cf8c9/Remote%20Habits/View/RHDashboardViewController.swift#L101-L102)

A refactor could prevent this duplicate code.

Remote Habits/View/RHSwitchWorkspaceViewController.swift
```diff
@@ -87,7 +98,21 @@ class RHSwitchWorkspaceViewController: RHBaseViewController, UITextFieldDelegate
         // Pass the selected object to the new view controller.
     }
     */
+    func routeToLogin() {
+        if let navController = self.parent as? UINavigationController, let presenter = navController.presentingViewController as? UINavigationController {
```
levibostian, line 102: What happens if this `if` is `false`? Seems error prone?

Remote Habits/View/RHSwitchWorkspaceViewController.swift
```diff
@@ -87,7 +98,21 @@ class RHSwitchWorkspaceViewController: RHBaseViewController, UITextFieldDelegate
...
+            let vc = main.instantiateViewController(withIdentifier: RHConstants.kLoginViewController) as! RHLoginViewController
+            presenter.setViewControllers([vc], animated: true)
+            self.dismiss(animated: true, completion: nil)
+        }
+    }
     
+    func textFieldShouldReturn(_ textField: UITextField) -> Bool {
+        self.view.endEditing(true)
+        return false
+    }
```
levibostian, line 114: Should this function go into a parent class or something? I imagine this code is very similar for many ViewControllers in this app.

Remote Habits/View/Xib/HabitDashboard/HabitTableViewCell.swift
```diff
@@ -30,7 +31,38 @@ class HabitTableViewCell: UITableViewCell {
         // Configure the view for the selected state
     }
     
+    func fillData() {
```
levibostian, line 34: Any reason for needing to call this function compared to:

```swift
var habitData: HabitData? {
  didSet {
    fillData()
  }
}
```

Remote Habits/View/Xib/HabitDetail/HabitDetailToggleTableViewCell.swift
```diff
@@ -12,6 +12,8 @@ class HabitDetailToggleTableViewCell: UITableViewCell {
     @IBOutlet weak var habitSwitch: UISwitch!
     @IBOutlet weak var habitTitle: UILabel!
     @IBOutlet weak var mainCellView: UIView!
+    var actionHandler : RHDashboardDetailActionHandler?
```
levibostian, line 15: Is this a delegate? Probably needs to be `weak`

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
```diff
@@ -18,11 +18,13 @@ class HabitReminderTableViewCell: UITableViewCell {
     @IBOutlet weak var headerTitle: UILabel!
     @IBOutlet weak var mainCellView: UIView!
     
+    var actionHandler : RHDashboardDetailTimeHandler?
```
levibostian, line 21: Is this a delegate? Probably needs to be `weak`

.sourcery.yml
```diff
@@ -0,0 +1,8 @@
+sources:
+    - "Remote Habits"
+templates:
+    - "Remote Habits/Templates"
+output:
+    "Remote Habits/autogenerated"
+args:
+    imports=CioTracking-CioMessagingPush
```
levibostian, line 8: I prefer that we put this configuration code into a `Makefile` like we do with the SDK https://github.com/customerio/customerio-ios/blob/664819caf48d23168ce97e027e51f8040b7922d1/Makefile#L8-L11

levibostian:
If you could, run `make lint` on the project and fix the lint issues. 

This process is automated for you with [git hooks setup](https://github.com/customerio/RemoteHabits-iOS/blob/main/docs/dev-notes/DEVELOPMENT.md)

Remote Habits/Repository/TrackRepository.swift
```diff
@@ -0,0 +1,24 @@
...
+import Foundation
+
+protocol TrackRepository {
+    func trackHabit(habitName: String, params : SelectedHabitData)
+}
+
+// sourcery: InjectRegister = "TrackRepository"
+class AppTrackRepository: TrackRepository {
+    private let cio: CustomerIO
+    private let cioErrorUtil: CustomerIOErrorUtil
```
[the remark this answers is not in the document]
ami-aman: File removed.

Remote Habits/Model/DashboardData.swift
```diff
@@ -39,3 +43,10 @@ struct HabitHeadersInfo : Hashable {
     let titleFontSize : Int?
     let titleFontName : String?
 }
+
+struct SelectedHabitData : Encodable {
```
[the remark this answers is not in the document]
ami-aman: SelectedHabitData has limited information like frequency ,starttime , endtime etc and updated information based on user selection whereas HabitDetail has all detail related to a habit (initial stub data).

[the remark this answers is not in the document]
ami-aman: Does this answer your question ? If yes, then please feel free to resolve conversation.

Remote Habits/Model/DashboardData.swift
ami-aman: I am going make modification into this in next task so this data model might look bit different with the next task. But thanks for letting me know about `AutoLenses`, it helps !

ami-aman: Directly consuming strings instead of `Date` or `Time`. Same thing can be done as a computed property but that would be double work, first changing from `Date/Time` to string in the controller to display the `Time` to the user in `textField` and another while fetching the value from the property

Remote Habits/Model/Profile.swift
ami-aman: Cred is short for Credential

Remote Habits/View/RHDashboardViewController.swift
ami-aman: Oops ! yes, this was removed, but as I updated the code this came over lol 
My bad !

ami-aman: Changes can be seen in latest commit.

Remote Habits/View/RHDashboardViewController.swift
ami-aman: User can land into Dashboard from either Login or Direct. This piece of code checks if `LoginViewController` is in the navigation stack, if yes then pops to the the controller else it adds it to stack and brings the user back to Login Screen. If you are thinking that we can directly push the user to LoginViewController then that would add two same controllers in nav stack which isn;t ideal. So this logic. I feel that you think that it is complex to understand, let me simplify this a bit. Please check latest commit for the changes.

Remote Habits/View/RHHabitDetailViewController.swift
ami-aman: Let me check this one with LJ and create a ticket for it. There are so many things we can add on at this stage of development.

Remote Habits/View/RHHabitDetailViewController.swift
ami-aman: Removed this code.

Remote Habits/View/RHHabitDetailViewController.swift
ami-aman: I am planning to move all this code to their respective `UITableViewCell` class. I have already done some part of it in `DashboardViewController` and will do the same for all controllers to keep the controllers light. But yes, to act on your comment, I am moving `selectionStyle` to it's respective class.

Remote Habits/View/RHHabitDetailViewController.swift
ami-aman: I am noting this down and will refactor in upcoming PRs. Already in process of making tweaks in data model.

Remote Habits/View/RHSwitchWorkspaceViewController.swift
ami-aman: There isn't going to be any use case where user is landing to this controller without a navigation controller. The only case that's possible is through deep linking and it is already marked as a TODO task in [this issue](https://github.com/customerio/RemoteHabits-iOS/issues/36). 

If this answers your question, please feel free to resolve the conversation.

Remote Habits/View/RHSwitchWorkspaceViewController.swift
ami-aman: I am not using this delegation method anywhere except `RHSwitchWorkspaceViewController`

Remote Habits/View/RHSwitchWorkspaceViewController.swift
```diff
@@ -97,4 +122,23 @@ class RHSwitchWorkspaceViewController: RHBaseViewController, UITextFieldDelegate
     @IBAction func apiKeyValueChanged(_ sender: Any) {
         didValueChange(sender, otherValue: siteIdInput.text ?? RHConstants.kEmptyValue)
     }
+    
+    @IBAction func switchWorkspaceButtonTapped(_ sender: UIButton) {
+        view.endEditing(true) // Close keyboard if it is open
+        showLoader()
```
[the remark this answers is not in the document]
ami-aman: Though `showLoader` is self explanatory but I do not mind changing it to `showLoadingView` for your convenience. Changes can be seen in this [commit](https://github.com/customerio/RemoteHabits-iOS/pull/35/commits/9ef2e6b1aea2b77197e4e4ddbcde283175e54fc8)

Remote Habits/View/Xib/HabitDashboard/HabitTableViewCell.swift
ami-aman: I doubt if what you are suggesting can populate the data in the view for `UITableViewCell`

Remote Habits/View/Xib/HabitDetail/HabitDetailToggleTableViewCell.swift
ami-aman: Yes it is a delegate but it's protocol a non-class protocol. Using non-class so that If needed, I could extend it to non-class types like struct. Right now we do not have any such use case but might come up in future.

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
ami-aman: Yes, that;s a delegate but using a non-class delegate so don't think it will create a strong reference. Reason why using a non-class protocol because it can be conformed by structs as well. No use case right now, but might use it in future.

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
```diff
@@ -31,4 +33,72 @@ class HabitReminderTableViewCell: UITableViewCell {
         // Configure the view for the selected state
     }
     
+    func setUpTextFieldsTimePicker() {
```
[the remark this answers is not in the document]
ami-aman: It is showing the same way on my device and simulator also. I thought it's new UI with updated OS. Let me check how it responds on older OS/device. I have iPhone 7, should check on that. 

Resolving this conversation, if I get any issue then will update with next PR. Noting it down.

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
```diff
@@ -31,4 +33,72 @@ class HabitReminderTableViewCell: UITableViewCell {
...
+        toTimeText.text = formatDateForDisplay(date: sender.date)
+        actionHandler?.updateTime(fromTime: fromTimeText.text ?? "",
+                                  toTime: toTimeText.text ?? "",
+                                  andFreq: Int(frequencyText.text ?? "0") ?? 0)
+    }
+    
+    func formatDateForDisplay(date: Date) -> String {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "h:mm a"
+        return formatter.string(from: date)
```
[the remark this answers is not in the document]
ami-aman: Moved to an extension. Changes committed. Hence resolving conversation

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
```diff
@@ -31,4 +33,72 @@ class HabitReminderTableViewCell: UITableViewCell {
...
+        let numberToolbar: UIToolbar = UIToolbar()
+        numberToolbar.barStyle = .default
+        numberToolbar.items = [UIBarButtonItem(title: "Done", style: UIBarButtonItem.Style.plain, target: self, action: #selector(hideKeyboard))]
+        numberToolbar.sizeToFit()
+        frequencyText.inputAccessoryView = numberToolbar
+        
+        fromTimeText.delegate = self
+        toTimeText.delegate = self
+        frequencyText.delegate = self
+        frequencyText.tag = 1001
```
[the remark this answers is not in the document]
ami-aman: Already have listed this as a TODO - to remove tag and use class instead. You will see the work around this in next few PRs. Thanks for keeping a check though.

Remote Habits/ViewModel/ProfileViewModel.swift
```diff
@@ -59,4 +59,29 @@ class ProfileViewModel: ObservableObject {
...
     }
+    
+    func validateWorkspace(forSiteId siteId: String, and apiKey : String, completion : @escaping (Bool) -> Void) {
+        profileRepository.validateWorkspace(forSiteId: siteId, and: apiKey) { (result: Result<ValidateWorkspaceResponse, HumanReadableError>) in
+         
+            DispatchQueue.main.async {
+                switch result {
+                case .success:
+                    do {
+                        let response = try result.get()
```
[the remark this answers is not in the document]
ami-aman: Done and committed.

Remote Habits/ViewModel/ProfileViewModel.swift
```diff
@@ -59,4 +59,29 @@ class ProfileViewModel: ObservableObject {
...
+                        guard let message = response.meta?.message, message.lowercased() == "nice credentials." else {
```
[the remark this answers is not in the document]
ami-aman: In the response, it is message body which returns nice credentials in case credentials are valid. Comapring to  make sure that message is valid and credentials are authorized.

ami-aman:
> 
Screenshot 2021-12-15 at 10 17 32 PM
I am trying to run make lint but getting this error. Any idea, why ? SwiftLint and SwiftFormat is already installed.

Remote Habits/Model/DashboardData.swift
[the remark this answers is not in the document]
levibostian: I am not a fan of maintaining 2 different structs that have almost identical overlap data in them. Is there a solution to avoid having to have 2 different structs? Can the code just use `HabitDetail` or maybe we can modify `SelectedHabitData` to contain `HabitDetail` in it like this?

```swift
struct SelectedHabitData {
  let habitDetail: HabitDetail
  // here, put any properties that are unique to `SelectedHabitData` that should not belong to `HabitDetail`
}
```

Remote Habits/Model/DashboardData.swift
levibostian: If we are using the `String` data type to display a `Date` in a human-readable format, I think that code should belong in the UI layer, not the data layer. 

So, I think it's better to have data structs like `DashboardData` to contain only `Date` and not `String` and then when the UI wants to display that date in a `textField`, it can create the `String` then:

```swift
textField.text = dashboardData.startTime.toString(format: .hourMinute12hour)

extension Date {
  func toString(format: DateTimeFormat) {
    ...
  }
}

enum DateTimeFormat {
  case hourMinute12hour

  var format: DateFormatter {
    swift self {
      case .hourMinute12Hour: return "h:mm a".dateFormatter
    }
  }
}
```

This is also nice because then every view in the UI gets to decide how to display the `Date` if they decide to display that data differently.

Remote Habits/ViewModel/ProfileViewModel.swift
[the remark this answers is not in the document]
levibostian: What would happen if the API ever changes the response string? Our code would break and all credentials would be marked as invalid in the app. 

I don't think that we need to deserialize the JSON response at all. If a 200 comes back from the API call, we know that the credentials are valid. No need to read the response JSON body at all.

Remote Habits/Model/Profile.swift
levibostian: I am not sure that this class or `ValidateWorkspaceResponse` should exist in the `Profile` file. 

But, we may not need these classes at all anyway [from this comment](https://github.com/customerio/RemoteHabits-iOS/pull/35#discussion_r771607644)

.sourcery.yml
levibostian: You can delete this file since all of the configuration that we need is contained inside of `Makefile` already.

levibostian:
If you install `swiftformat` via `brew install swiftformat` on your computer, that error should go away and you can lint again. 

Sorry, I tried to make this instruction to install `swiftformat` convenient by [printing a message to you](https://github.com/customerio/RemoteHabits-iOS/blob/main/hooks/post-install.sh#L12) after git hooks are installed. That might not be the best place to put those instructions. Maybe they should exist just in the [development doc](https://github.com/customerio/RemoteHabits-iOS/blob/main/docs/dev-notes/DEVELOPMENT.md#development) instead.

Remote Habits/View/RHDashboardViewController.swift
levibostian: I would like to try to find some code that's less boilerplate. I am wondering if there is a more simple way that we can use across more screens of the app. 

If the app user decides to logout, I think we can destroy all of the app's UI and take them to the login screen just like if the app opened for the first time after install, right? 

If so, can we instead do something like this?

```swift 
class XViewController {
  func logout() {
    // delete all data 
    let sceneDelegate = ...
    sceneDelegate.setRootScreen(.login)
  }
}

class SceneDelegate {
  func setRootScreen(_ screen: RootScreenType) {
    window.rootViewControler = screen.viewController
    window.makeKeyAndVisible()
  }

  enum RootScreenType {
    case dashboard
    case login

    var viewController: ViewController {
       ...
     }
  }
}
```

That's my one idea. Is there a reason we need to keep the UI hierarchy when logging out on this screen?

Remote Habits/View/Xib/HabitDashboard/HabitTableViewCell.swift
levibostian: I like the solution of automatically altering the cell views when you set data on the cell. That's why I suggested the `didSet {}` solution. I like that idea compared to having to always remember to call `fillData()` when the cell can do that for us. 

I am not understanding why the `didSet {}` solution wouldn't populate the views in the cell?

Remote Habits/Model/DashboardData.swift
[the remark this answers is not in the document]
ami-aman: I might have to change this with the new requirement of retaining user data. If not, then let me see how I can optimise this better so that we both come to same page !

Remote Habits/Model/DashboardData.swift
ami-aman: I will make these changes with other PR because it will require multiple file changes. But i get your point here.

Remote Habits/ViewModel/ProfileViewModel.swift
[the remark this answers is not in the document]
ami-aman: Let me check this with the backend team and understand the possibility of having another message with status 200. It happened with me previously that with status 200 the api returned a readable error message in the response. But yes, this I should check with the backend regarding it. Thanks.

.sourcery.yml
ami-aman: Cool. Let me delete it and update the PR.

ami-aman:
I will raise another PR with lint fixes. Will create single PR for Lint fixes + removing headers in files.

PR #39, opened 2021-12-17
ci: add swift lint and PR title checks

levibostian:
Note: These CI checks are enabled on the iOS SDK today. This adds it to the RH project. 

Adding lint checks to CI server to:
1. Help remind us to lint the Swift code when we make PRs and fix those errors. The CI will simply fail the build reminding you to fit the lint errors on your own machine and push the fixes up. This can be avoided by installing the [git hooks](https://github.com/customerio/RemoteHabits-iOS/blob/main/docs/dev-notes/DEVELOPMENT.md#development) for the project. 
2. Make sure our PR title aligns with [format](https://www.conventionalcommits.org/en/) that CI requires to make a successful deployment. I noticed that when I was out on PTO, a commit message was pushed to `main` for deployment that had a typo in it. This resulted in a deployment not being made when it should have been. By being able to lint the PR title before it's merged in, this helps us to have a higher chance of success deploying the code. 

---

The `main` branch has not had `make format && make lint` run on it recently which made this PR longer then it needed to be. To help with reviewing this PR, I made comments in the PR where I mention the big changes that this PR made.

PR #42, opened 2021-12-29
feat: retain user data on dashboard 

ami-aman:
- Retain user selection in habits dashboard
- Retain user selection in dashboard habit detail
- Changed data model to support persistent data using CoreData
- Addressed PR comments
- Changed the way user selects time in habit detail screen
- Reset data on logout
- Update dashboard when some habit is enabled/disabled from a specific detail screen
- Minor self analysed bug fixes
- app icons for other dimensions


Remote Habits/View/Common/RHBaseViewController.swift
```diff
@@ -7,9 +7,17 @@
...
 
+
+enum ControllerSource {
+    case habitdashboard
+    case habitdetail
+}
 class RHBaseViewController: UIViewController {
 
     var loaderView : UIView?
+    let habitsDataManager = HabitDataManager()
```
levibostian, line 18: Reason why the `HabitDataManager` isn't being injected via the `DI.shared` graph like the `TrackerViewModel` is below?

Remote Habits/View/Common/RHBaseViewController.swift
```diff
@@ -121,6 +129,15 @@ class RHBaseViewController: UIViewController {
         loaderView?.removeFromSuperview()
     }
 
+    func updateHabit(forActivity activity: String?, selectedHabit: SelectedHabitData, andSource source : ControllerSource) {
```
levibostian, line 132: Hmmmm... 🤔 

This looks like business logic to me. Should we move this code into a ViewModel class instead?

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -13,55 +13,72 @@ class RHDashboardViewController: RHBaseViewController {
     @IBOutlet weak var dashboardTableView: UITableView!
     
     // MARK: - --VARIABLES--
-    var dashboardData  = UserHabit()
+    let remoteHabitsData = RemoteHabitsData()
```
levibostian, line 16: Reason why the `HabitDataManager` isn't being injected via the DI.shared graph?

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -13,55 +13,72 @@ class RHDashboardViewController: RHBaseViewController {
...
-        dashboardTableView.dataSource = self
         // Do any additional setup after loading the view.
-        
-        
     }
     
     // MARK: - --FUNCTIONS--
-    func setupDashboardTableView() {
+    func addNotifierObserver() {
+        NotificationCenter.default.addObserver(self, selector: #selector(self.reloadHabitsData(notification:)), name: Notification.Name(RHConstants.kHabitsUpdatedIdentifier), object: nil)
```
levibostian, line 36: We need to also remove the observer when the class `deinit` is called.

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -13,55 +13,72 @@ class RHDashboardViewController: RHBaseViewController {
...
-        dashboardData = RHStubData().getStubData()
+        dashboardTableView.reloadData()
+    }
+    
+    func setupDashboardTableView() {
         
-        // HabitTableViewCell move to constant file
         dashboardTableView.register(UINib(nibName: RHConstants.kHabitTableViewCell, bundle: nil), forCellReuseIdentifier: RHConstants.kHabitTableViewCell)
         dashboardTableView.rowHeight = UITableView.automaticDimension
         dashboardTableView.estimatedRowHeight = 80
```
levibostian, line 48: I still think an extension function would be handy here. What do you think? 

```swift
dashboardTableView.setAutomaticRowHeight()
```

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -13,55 +13,72 @@ class RHDashboardViewController: RHBaseViewController {
...
+            navigateToWorkspace()
+        default:
+            break
+        }
     }
     
-    func route(withData : HabitData) {
-        if let viewController  = UIStoryboard(name: RHConstants.kStoryboardMain, bundle: nil).instantiateViewController(withIdentifier: RHConstants.kHabitDetailViewController) as? RHHabitDetailViewController {
+    func navigateToDashboardDetail(withData : Habits?) {
+        if let viewController  = UIStoryboard(name: RHConstants.kStoryboardMain, bundle: nil).instantiateViewController(withIdentifier: RHConstants.kHabitDetailViewController) as? RHHabitDetailViewController, let habitData = withData {
```
levibostian, line 67: The `UIStoryboard` is boilerplate heavy. How about an extension to make the code look more like this:

```swift
let habitDetailViewController: RHHabitDetailViewController? = ViewController.fromStoryboard(.main, identifier: RHConstants.kHabitDetailViewController)



extension UIStoryboard {
  static func fromStoryboard<VC: UIViewController>(_ storyboard: StoryboardId, identifier: String) -> VC? {
    UIStoryboard(name: storyboard.rawValue, bundle: nil).instantiateViewController(withIdentifier: identifier) as? VC
  }
```

I think that reads better. What do you think? 

We could also make the `RHConstants.kHabitDetailViewController` into a String enum. And default the storyboad name since there is only 1 storyboard in this project. 

```swift
let habitDetailViewController: RHHabitDetailViewController? = ViewController.fromStoryboard(identifier: .habitDetail)
```

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -73,15 +90,15 @@ extension RHDashboardViewController : UITableViewDelegate {
...
+        let headerData = dashboardHeaders[section]
         let label = UILabel()
         label.frame = CGRect.init(x: 16, y: 5, width: headerView.frame.width-20, height: headerView.frame.height-10)
         if section != 0 {
             label.frame = CGRect.init(x: 16, y: 60, width: headerView.frame.width-20, height: 26)
         }
         
-        label.text = headerData?.headerTitle
-        label.font = UIFont(name: headerData?.titleFontName ?? RHConstants.kEmptyValue, size: CGFloat(headerData?.titleFontSize ?? 17))
+        label.text = headerData.headerTitle ?? RHConstants.kEmptyValue
```
levibostian, line 100:
```suggestion
        label.text = headerData.headerTitle ?? ""
```

I think this reads better. I don't think we need a constant for empty string as it makes the code harder to read and doesn't add much value.

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -73,15 +90,15 @@ extension RHDashboardViewController : UITableViewDelegate {
     func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
         let headerView = UIView.init(frame: CGRect.init(x: 0, y: 0, width: tableView.frame.width, height: 97))
         headerView.backgroundColor = self.view.backgroundColor
-        let headerData = dashboardData[section].first?.key
+        let headerData = dashboardHeaders[section]
         let label = UILabel()
         label.frame = CGRect.init(x: 16, y: 5, width: headerView.frame.width-20, height: headerView.frame.height-10)
         if section != 0 {
             label.frame = CGRect.init(x: 16, y: 60, width: headerView.frame.width-20, height: 26)
         }
```
levibostian, line 98: If we instead design the table view header in the Storyboard, wouldn't we be able to get rid of this boilerplate code here?

Remote Habits/Model/DashboardData.swift
```diff
@@ -7,46 +7,47 @@
...
-    var actionButtonValue : String?
-    var actionType : HabitActionType?
-}
-
-
 struct HabitHeadersInfo : Hashable {
     let headerTitle : String?
     let titleFontSize : Int?
     let titleFontName : String?
+    let ids : [Int]?
```
levibostian, line 14: Identifiers for what? Maybe a property name that's more descriptive would be better here.

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -93,38 +110,35 @@ extension RHDashboardViewController : UITableViewDelegate {
         
         if indexPath.section != 0 { return }
         
-        guard let habitData = dashboardData[indexPath.section].first?.value[indexPath.row], let habitName = habitData.title else {
-            //Show error
-
+        guard let id = dashboardHeaders[indexPath.section].ids?[indexPath.row], let habitData = habitsDataManager.getHabit(forIds: [id])?.first else {
```
levibostian, line 113: `habitsDataManager.getHabit(forIds: [id])?.first`

Couple questions about this function. 

1. The function name is `getHabit()` but the result is plural, not singular. I think `getHabits()` would be better. 
2. Instead of `getHabits(forIds: [id])?.first`, couldn't you make another function in `HabitsDataManager` and have it be `getHabit(id])?`

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -93,38 +110,35 @@ extension RHDashboardViewController : UITableViewDelegate {
...
             return
         }
         
-        let selectedHabit = SelectedHabitData(title: habitName, frequency: habitData.habitDetail?.frequency, startTime: habitData.habitDetail?.startTime, endTime: habitData.habitDetail?.endTime)
-        trackerViewModel.trackHabitActivity(withName: RHConstants.kHabitClicked, forHabit: selectedHabit)
-        self.route(withData : habitData)
+        let selectedHabit = SelectedHabitData(title: habitData.title, frequency: Int(habitData.frequency), startTime: habitData.startTime?.formatDateToString(inFormat: .time12Hour), endTime: habitData.endTime?.formatDateToString(inFormat: .time12Hour), id: Int(habitData.id), isEnabled: habitData.isEnabled)
```
levibostian, line 117: Why are we constructing a new object `SelectedHabitData`? Can't we just use `HabitData` instead? I don't see a point in having a class *just* for when a piece of data is selected vs not selected.

Remote Habits/View/RHHabitDetailViewController.swift
```diff
@@ -56,6 +55,10 @@ class RHHabitDetailViewController: RHBaseViewController {
         }
     }
     
+    func updateHabits(with selectedHabit : SelectedHabitData) {
+        habitsDataManager.updateHabit(withData: selectedHabit)
+        NotificationCenter.default.post(name: Notification.Name(RHConstants.kHabitsUpdatedIdentifier), object: nil)
```
levibostian, line 60: After reading this code, I am starting to see a pattern. There is a lot of code in the ViewControllers that I would consider to be business-level logic. Sure, the code above is calling `habitsDataManager.updateHabit(withData: selectedHabit)` to update the CoreData database instead of having the CoreData code directly in the ViewController code but I do believe that these 2 functions above should be coupled together into a re-useable function. 

Something like:

```swift
class HabitViewModel {

  func updateHabit(_ habit: HabitData) {
    habitsDataManager.updateHabit(withData: selectedHabit)
    NotificationCenter.default.post(name: Notification.Name(RHConstants.kHabitsUpdatedIdentifier), object: nil)
  } 
}

class RHHabitDetailViewController: UIViewController {
  ...
  habitViewModel.updateHabit(habit)
  ...
}
```

What if you had another ViewController class that needed to update habits? As the code stands today, you would copy/paste the above 2 functions. 

By having a class like a ViewModel between the business layer and View layer, any ViewController can re-use this function preventing the copy/paste behavior.

Remote Habits/View/RHLoginViewController.swift
```diff
@@ -21,6 +21,7 @@ class RHLoginViewController: RHBaseViewController {
     
     // MARK: - --VARIABLES--
     var userManager = DI.shared.userManager
+//    let habitsDataManager = HabitDataManager()
```
levibostian, line 24: Remove this.

Remote Habits/View/RHLoginViewController.swift
```diff
@@ -75,7 +76,6 @@ class RHLoginViewController: RHBaseViewController {
                 field.iconText = "\u{f007}"
```
levibostian, line 76: https://github.com/customerio/RemoteHabits-iOS/pull/31/files/d16c81cf9e3e6b1c9293fec9cfce7849a3159834#r766083085

Remote Habits/View/RHLoginViewController.swift
```diff
@@ -139,8 +145,8 @@ class RHLoginViewController: RHBaseViewController {
                 }
                 else {
                     floatingLabelTextField.errorMessage = RHConstants.kEmptyValue
+                    floatingLabelTextField.text = floatingLabelTextField.text
```
levibostian, line 148: What is the point of this code?

Remote Habits/View/RHLoginViewController.swift
```diff
@@ -122,13 +122,19 @@ class RHLoginViewController: RHBaseViewController {
...
         loginButtonState()
         
-        userManager.isGuestLogin = isGuestLogin
         if let viewController  = UIStoryboard(name: RHConstants.kStoryboardMain, bundle: nil).instantiateViewController(withIdentifier: RHConstants.kDashboardViewController) as? RHDashboardViewController {
             viewController.isSourceLogin = true
             navigationController?.pushViewController(viewController, animated: true)
         }
     }
     
+    func fillHabitData() {
```
levibostian, line 131:
```suggestion
    func populateInitialHabitData() {
```

I think a more descriptive function name would be good to use here. `fillHabitData` doesn't describe what the function is doing.

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
```diff
@@ -33,64 +35,75 @@ class HabitReminderTableViewCell: UITableViewCell, UITextFieldDelegate {
...
+                if textField == toTimeText {
+                    selector = #selector(updateEndTime(_:))
+                }
+                datePickerST.addTarget(self, action: selector, for: .valueChanged)
+                datePickerST.preferredDatePickerStyle = .wheels
+                textField?.inputView = datePickerST
+            }
+            
+            // Done button accessory view for all textfields
+            let doneToolbar: UIToolbar = UIToolbar(frame: CGRect(x: 0, y: 0, width: UIScreen.main.bounds.width, height: 35))
```
levibostian, line 64: I think it would be a good idea to make the `35` a constant to make it reuseable in the code. 

Is the height set so that it allows 1 row of a UILabel to fit inside of the toolbar? If so, maybe something like this would be useful:

```swift
UIToolbar.getSingleRowToolbar()
```

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
```diff
@@ -33,64 +35,75 @@ class HabitReminderTableViewCell: UITableViewCell, UITextFieldDelegate {
...
     @objc func updateEndTime(_ sender : UIDatePicker) {
         toTimeText.text = formatDateForDisplay(date: sender.date)
-        actionHandler?.updateTime(fromTime: fromTimeText.text ?? "",
-                                  toTime: toTimeText.text ?? "",
-                                  andFreq: Int(frequencyText.text ?? "0") ?? 0)
+        updateHabitTime()
+    }
+    
+    func updateHabitTime(withFreq freq : Int? = nil) {
+        let frequency = freq ?? Int(frequencyText.text ?? "0")
```
levibostian, line 90: This function takes in a parameter but doesn't use it.

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
```diff
@@ -33,64 +35,75 @@ class HabitReminderTableViewCell: UITableViewCell, UITextFieldDelegate {
...
+        actionHandler?.updateTime(with: selectedHabit)
     }
     
     func formatDateForDisplay(date: Date) -> String {
         return date.formatDateToString(inFormat: .time12Hour)
     }
     
     func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
-        return textField.tag == 1001 ? true : false
+        return ((textField as? FrequencyTextField) != nil) ? true : false
```
levibostian, line 101:
```suggestion
        return textField is FrequencyTextField
```

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
```diff
@@ -33,64 +35,75 @@ class HabitReminderTableViewCell: UITableViewCell, UITextFieldDelegate {
...
     }
 
     func textFieldDidEndEditing(_ textField: UITextField) {
-        if textField.tag == 1001 {
-            actionHandler?.updateTime(fromTime: fromTimeText.text ?? "",
-                                      toTime: toTimeText.text ?? "",
-                                      andFreq: Int(textField.text ?? "0") ?? 0)
+        if ((textField as? FrequencyTextField) != nil) {
```
levibostian, line 105:
```suggestion
        if textField is FrequencyTextField {
```

Remote Habits/Extension/StringExtensions.swift
```diff
@@ -22,6 +22,15 @@ extension String {
     static func random(length: Int = 10) -> String {
         String((0 ..< length).map { _ in abcLetters.randomElement()! })
     }
+    
+    func stringToDate(withFormat format: DateFormat)-> Date?{
```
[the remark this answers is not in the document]
ami-aman: Made changes, can be found in recent commit

Remote Habits/Model/DashboardData.swift
```diff
@@ -7,46 +7,47 @@
...
 }
 
 struct SelectedHabitData : Encodable {
     let title : String?
     let frequency : Int?
     let startTime : String?
     let endTime : String?
+    let id : Int
+    let isEnabled : Bool?
+}
```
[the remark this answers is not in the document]
ami-aman: 1. This is not related to Core data. 
2. It is always better and safe to use optionals to avoid unexpected crashes
3. We need to keep them optional because these properties can either be updated from dashboard or detail screen and might contain the default values and the default values are optional values

Remote Habits/Stub/RHStubData.swift
```diff
@@ -6,45 +6,120 @@
...
     
-    var dashboardData : UserHabit = UserHabit()
     let userManager = DI.shared.userManager
     
-    func getStubData() -> [[HabitHeadersInfo : [HabitData]]]{
-        let isLoggedIn = !(userManager.isGuestLogin ?? true)
+    func getHabitHeaders() -> [HabitHeadersInfo] {
         
-        let habitInfo_hydration = HabitDetail(isHabitEnabled: false, frequency: 8, startTime: "9:00 AM", endTime: "5:00 PM", description: "Drinking enough water each day is crucial for many reasons: to regulate body temperature, keep joints lubricated, prevent infections, deliver nutrients to cells, and keep organs functioning properly. Being well-hydrated also improves sleep quality, cognition, and mood.", actionButtonValue: nil, actionType: nil)
+        let isLoggedIn = !(userManager.isGuestLogin ?? true)
```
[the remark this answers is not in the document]
ami-aman: I always prefer using optionals to keep the code safe and crash free. 
Also, isGuestLogin holds no value until unless user either logs in or uses guest login button to enter into the app. and it's value gets nil when user logs out, fills only when either of the two processes happen.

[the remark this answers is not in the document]
ami-aman: Not using `userManager.isGuestLogin` any longer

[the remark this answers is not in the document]
ami-aman: I added `isGuestLogin` because with `isLoggedIn` value was based on email. Made few tweaks and now solely using `userManager.isLoggedIn` and completely removed isGuestLogin

Remote Habits/Stub/RHStubData.swift
```diff
@@ -6,45 +6,120 @@
...
+        let title = isLoggedIn ? userManager.userName : "Guest"
+        let section_first = HabitHeadersInfo(headerTitle: "\(title ?? "Your")'s Habits", titleFontSize: 34, titleFontName: "SFProDisplay-Bold", ids: [1,2,3])
```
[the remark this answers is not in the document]
ami-aman: `HabitHeadersInfo` has dynamic information based on user's type of login, so placed in data layer than UI layer.

Remote Habits/Stub/RHStubData.swift
```diff
@@ -6,45 +6,120 @@
...
-        let section1 = HabitHeadersInfo(headerTitle: "\(title ?? "Your")'s Habits", titleFontSize: 34, titleFontName: "SFProDisplay-Bold")
-        let section2 = HabitHeadersInfo(headerTitle: "Details", titleFontSize: 17, titleFontName: "SFProDisplay-Bold")
-        dashboardData.append([section1 : habitData])
-        dashboardData.append([section2 : habitDetail])
-        return dashboardData
+        return [userData, workspaceData, sdkData]
+    }
+    
+    func getUserData() -> HabitsData {
+        let isLoggedIn = !(userManager.isGuestLogin ?? true)
```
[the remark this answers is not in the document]
ami-aman: Made changed

Remote Habits/Stub/RHStubData.swift
```diff
@@ -6,45 +6,120 @@
...
+                           startTime: nil,
+                           endTime: nil,
+                           habitDescription: nil,
+                           actionName: isLoggedIn ? "Log Out" : "Log In",
+                           actionType: isLoggedIn ? .logout : .login)
     }
 }
+/**
+ 
+ fresh login ->
```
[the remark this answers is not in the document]
ami-aman: Oops ! removed

Remote Habits/View/Common/RHBaseViewController.swift
ami-aman: I tried it but was facing some issue. Let;s connect to get this done.

Remote Habits/View/RHDashboardViewController.swift
ami-aman: I tried it but was facing some issue. Let;s connect to get this done.

Remote Habits/Stub/RHStubData.swift
```diff
@@ -6,45 +6,120 @@
 //
 
 import Foundation
-import CioTracking
 
-typealias UserHabit = [[HabitHeadersInfo : [HabitData]]]
-
-class RHStubData {
+class RemoteHabitsData {
```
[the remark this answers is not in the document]
ami-aman: There are `n `number of ways to do one thing. Anything is possible but doing so will put massive load on view controller and make it heavy with lots of conditions to match and show.

Remote Habits/Stub/RHStubData.swift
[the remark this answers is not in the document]
levibostian: From my understanding, if `UserManager.email != nil`, then a profile has been identified in the app. 

If `UserManager.email == nil && UserManager.username != nil`, that's a guest login. Correct? 

If the app allows guest logins (I am not sure if "guest" is the best name for this because in the CIO docs we refer to this as "anonymous"), we shouldn't remove the `isGuestLogin` property?

Remote Habits/Stub/RHStubData.swift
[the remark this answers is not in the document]
levibostian: But this dynamic information can be generated in the UI layer instead of here in the business layer. 

The title + font size + font are all UI related pieces of information. They should exist only in the UI layer.

Remote Habits/Model/DashboardData.swift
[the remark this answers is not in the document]
levibostian: > "and the default values are optional values"

Why are the default values optional values? 

After [discussing if we need for this class](https://github.com/customerio/RemoteHabits-iOS/pull/42#discussion_r777710472) we can decide on this.

Remote Habits/Stub/RHStubData.swift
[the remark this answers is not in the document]
levibostian: After re-reading my comment, I can see how all of the properties are UI related code - not business logic + UI related. 

I should have reworded myself when I said, "move it all into the ViewController..." because I agree that bloated ViewControllers are not a good idea either. 

I do believe there is a better implementation that can be done here. Use of custom Views maybe? I will have to think about it a bit. 

I don't see this a as a deal breaker so we can come back to it later.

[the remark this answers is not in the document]
levibostian: I would appreciate updating the name of the File since the name of the class changed.

Remote Habits/View/RHDashboardViewController.swift
levibostian: I came up with a possible solution in [a commit](https://github.com/customerio/RemoteHabits-iOS/pull/39/commits/a4a5a88df24be0d5cf121bbab81d1e18493279cd) for fixing lint errors. 

The solution I came up with replaces this code:

```swift
if let viewController  = UIStoryboard(name: RHConstants.kStoryboardMain, bundle: nil).instantiateViewController(withIdentifier: RHConstants.kHabitDetailViewController) as? RHHabitDetailViewController
 ```

with this code:
```swift
let viewController  = RHHabitDetailViewController.newInstance()
```

If we like the new API ^^^ then we can skip this suggestion in this PR and it will get fixed when we merge this PR.

Remote Habits/Model/DashboardData.swift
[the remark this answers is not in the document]
ami-aman: Let me think about all possible scenarios and make the changes to non-optionals if needed. But I would like to mention that I will be making these changes in a separate PR and not in this one. Reasons - 1. Delay in PR getting merged and rolled out. 2. Re-review of all files. (Changes will have to be done in most of the files). Hope that's fine.

Remote Habits/Stub/RHStubData.swift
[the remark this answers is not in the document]
ami-aman: I like to do this because it helps in easy implementation like in future if we need to add one more header, all we would need to do is input detail in model and rest will be handled by controller. That's how experience services work. I am okay to change this to traditional way but this provides more flexibility

Remote Habits/Stub/RHStubData.swift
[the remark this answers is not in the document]
ami-aman: Yes, I will be renaming few other files with this one in the lint PR. This is a TODO task for me. There are other files to be renamed too, like ones starting with RH.

Remote Habits/View/Common/RHBaseViewController.swift
ami-aman: We can. Not a big deal. This function updates values in core data and also track activities. Putting this code in either Profile or Tracker VM would mean mixing up both the things. But let me see if I can still manage to move this to a separate file and keep it independent.

Remote Habits/View/RHDashboardViewController.swift
ami-aman: Oops ! Thanks for pointing out this major miss. I have made the changes

Remote Habits/View/RHDashboardViewController.swift
ami-aman: Done.

Remote Habits/View/RHDashboardViewController.swift
ami-aman: This seems reasonable. Thanks for the suggestion

Remote Habits/View/RHDashboardViewController.swift
ami-aman: Hmm... I do not see any issue writing up the code to design section header. 🤔🤔

Remote Habits/Model/DashboardData.swift
ami-aman: Ids for which all items to be shown to the user, For example, right now we are showing ids [4,5,6] to the user but I discussed with LJ that for now we have to remove id 6 so ids would be [4,5] and rest of the work will be handled by the code. All i have to do is remove an id and not any specific piece of code.

Remote Habits/View/RHDashboardViewController.swift
ami-aman: hehe formerly this function was providing one habit at a time. Thanks for calling this out.

Remote Habits/View/RHHabitDetailViewController.swift
ami-aman: This function has been removed in the latest commit. Not needed.

Remote Habits/View/RHLoginViewController.swift
ami-aman: https://github.com/Skyscanner/SkyFloatingLabelTextField/issues/334

ami-aman: This is an open issue with the library and it seems to be an iOS issue which has this workaround

Remote Habits/View/RHLoginViewController.swift
ami-aman: Done.

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
ami-aman: Currently it is just done button in the toolbar and I do not think we are ever going to add any other button to the toolbar. But yes, have moved 35 to constants

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
ami-aman: But I see it being used 🤔

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
ami-aman: Done

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
ami-aman: ✔️

Remote Habits/View/RHDashboardViewController.swift
ami-aman: Not using HabitData anymore

Remote Habits/View/RHLoginViewController.swift
ami-aman: ✔️

Remote Habits/Manager/DataManager.swift
```diff
@@ -0,0 +1,131 @@
+import Foundation
```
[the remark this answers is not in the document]
ami-aman: 👍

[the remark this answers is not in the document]
ami-aman: Let me also work around to find an alternate if you feel its heavy

Remote Habits/Model/DashboardData.swift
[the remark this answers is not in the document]
levibostian: I am OK with changes being done in a future PR as long as a GitHub issue is created discussing the work to be done. With this in mind, I think it's safe to say that small changes to a code base are worth performing the changes in the same PR. Bigger refactors are worth making a separate issue for and doing in the future.

Remote Habits/View/RHDashboardViewController.swift
levibostian: I like to use the Storyboard/xib as much as reasonable if we are to commit to using Storyboard in a project. 

If we are designing UITableViewCell using storyboard/xib, why not also use storyboard/xib for headers, too? Being consistent is what I am going for with this comment.

Remote Habits/Model/DashboardData.swift
levibostian: "ids" are not descriptive to me. By saying, "we are showing ids [4,5,6]", that does not explain what the code is actually doing. What is id 4? What is id 6? I do understand what you mean by how using an array of `Int` is convenient because if you want to edit the headers of the UI, you just need to modify the array of `Int` and the UI changes. However, from someone glancing at this code, it's difficult to understand what the code is doing with random `Int`s. 

Could we replace the numbers 4,5,6 with an enum maybe? Would that improve the readability of the code? 

```swift
HabitHeadersInfo(..., ids: [HabitHeader.Focus])
```

Remote Habits/View/RHLoginViewController.swift
levibostian: Ok. that's understandable. Mind adding a comment above the line of code explaining that the code is a workaround? And a link to the GH issue in the comment would be handy, too.

Remote Habits/View/Xib/HabitDetail/HabitReminderTableViewCell.swift
levibostian: 🤦‍♂️ yeah, you're right. `freq` is right there in front of my face.

Remote Habits/Common/RHConstants.swift
```diff
@@ -43,4 +43,7 @@ struct RHConstants {
     static let kHabitClicked = "clicked_habit"
     static let kHabitEnabled = "habit_enabled"
     static let kHabitDisabled = "habit_disabled"
+    
+    // Toolbar
+    static let kToolBarHeight = 35.0
```
levibostian, line 48: I would prefer a variable that's more descriptive because maybe not all Toolbars will want to be this height? 

What about using a variable such as `kToolbarSingleRowHeight` implying that if you want to have a Toolbar tall enough for 1 row of Views in it, use this height?

Remote Habits/View/RHDashboardViewController.swift
```diff
@@ -31,6 +31,11 @@ class RHDashboardViewController: RHBaseViewController {
         // Do any additional setup after loading the view.
     }
     
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        NotificationCenter.default.removeObserver(self)
```
levibostian, line 36: Is this the correct function to remove the observer? 

I ask because if we are removing the observer with `viewWillDisappear()`, then I think we should be using `viewWillAppear()` for adding the observer, right?

Remote Habits/Model/DashboardData.swift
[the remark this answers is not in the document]
ami-aman: Cool. I will create a github ticket to track this.

Remote Habits/Common/RHConstants.swift
ami-aman: Can there be any scenario where we might have a toolbar with other height other than having a done button or a cancel button ? 🤔 I assume, no!

Remote Habits/View/RHDashboardViewController.swift
ami-aman: done

Remote Habits/Model/DashboardData.swift
ami-aman: I can make this change in the upcoming PR since making change within this PR will cause conflicts. Once merged I will create a PR for updating the data models

Remote Habits/Stub/RHStubData.swift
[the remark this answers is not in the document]
levibostian: I am still not the biggest fan of it but it's not a reason to block this PR. I'll resolve it.

Remote Habits/Common/RHConstants.swift
levibostian: Thinking about this again, I think it would be better to have a custom View `UIToolbar` subclass instead of a constant for the height. 

Why? Every time that you create a `UIToolbar` instance in the app, you need to remember to use this constant `kToolbarHeight` value. If you forget, it creates tech debt or a UI bug. Instead, if we had a custom View class, you don't need to remember to do anything except create an instance. 

Any other ideas?

Remote Habits/View/RHLoginViewController.swift
```diff
@@ -8,6 +8,10 @@
 import UIKit
 import SkyFloatingLabelTextField
 
+enum Icons: String {
```
levibostian, line 4: Since these icons can be used anywhere in the app, they should be moved into a separate file except in this file?

Remote Habits/View/RHDashboardViewController.swift
levibostian: I could be wrong, but there are scenarios when `viewWillAppear()` will be called multiple times for a class compared to `viewDidLoad()` which only gets called once. 

Would it better to add the observer in `viewDidLoad()` and remove the observer in `deinit`? This has a higher chance of only calling add and remove once.

ami-aman: 👍

Remote Habits/Common/RHConstants.swift
ami-aman: Agreed. I have created a custom class

Remote Habits/View/RHLoginViewController.swift
ami-aman: Done

Remote Habits/Common/RHConstants.swift
ami-aman: This custom class will also allow multiple buttons in same toolbar, all we need to do is pass enum in the parameter.

levibostian: Great change 👍! I find this new custom class less error prone.

Remote Habits/Manager/DataManager.swift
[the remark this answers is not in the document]
levibostian: This is something that can be done at a later time. No need to stop this PR.

levibostian (approved):

PR #44, opened 2022-01-03
feat: add dark mode support

ami-aman:
This PR provides support for dark mode in Remote Habits app. This was one of the major feedback from multiple users

IMG_1316


ami-aman:
This feature branch is branched from the previous PR (branch -> RH-retain-habits) so this PR has all commits from the previous PR. Would suggest to view only last 2 commits. 

Dark Mode - [5bb307a](https://github.com/customerio/RemoteHabits-iOS/pull/44/commits/5bb307a9d58d62879c7060bcdef5c39d5204d906) 
Minor Bug fix - [7160524](https://github.com/customerio/RemoteHabits-iOS/pull/44/commits/71605248819b719fa28b19b549dea0cb184ee951)

Remote Habits/Extension/ViewExtension.swift
```diff
@@ -1,14 +1,14 @@
...
+    func setCornerRadius() {
+        self.layer.cornerRadius = 13
     }
-}
-
-extension Binding {
-    static func get(_ getValue: @escaping () -> Value) -> Binding<Value> {
-        Binding(get: getValue, set: { _ in })
+    
+    func customiseView() {
```
levibostian, line 10: I don't like this function. It's an extension on `UIView` which means that *all* `UIView`s in this whole app *should* have a corner radius and a white background. 

After looking at this pull request to see how it's being used, I think it's best to remove this function all together. Instead of the `UIView`s calling this function, have 2 lines of code to set the radius and set the background color instead of calling this function.

Remote Habits/View/Common/RHBaseViewController.swift
```diff
@@ -102,8 +102,8 @@ class RHBaseViewController: UIViewController {
         if #available(iOS 13.0, *) {
             let navBarAppearance = UINavigationBarAppearance()
             navBarAppearance.configureWithOpaqueBackground()
-            navBarAppearance.titleTextAttributes = [.foregroundColor: titleColor]
-            navBarAppearance.backgroundColor = backgoundColor
+            navBarAppearance.titleTextAttributes = [.foregroundColor: self.traitCollection.userInterfaceStyle == .dark ? RHColor.LabelBlack : titleColor]
+            navBarAppearance.backgroundColor = self.traitCollection.userInterfaceStyle == .dark ? RHColor.DefaultBackground : backgoundColor
```
levibostian: Instead of referring to `self.traitCollection.userInterfaceStyle == .dark`, couldn't we create an Xcode color asset for the navigation bar title text color and the background color and then do this:

```swift
navBarAppearance.titleTextAttributes = [.foregroundColor: RHColor.navigationTitleTextColor]
navBarAppearance.backgroundColor = RHColor.navigationBarBackgroundColor
```

Remote Habits/View/RHLoginViewController.swift
```diff
@@ -80,6 +80,8 @@ class RHLoginViewController: RHBaseViewController {
                 break
             }
             
+            field.placeholderColor = RHColor.LabelGray
+            field.textColor = UIColor.black
```
levibostian, line 84: Avoid using `UIColor` as it will not be dark mode friendly. Use a `RHColor` instead.

Remote Habits/View/Xib/HabitDashboard/HabitTableViewCell.swift
```diff
@@ -22,12 +22,16 @@ class HabitTableViewCell: UITableViewCell {
     override func awakeFromNib() {
         super.awakeFromNib()
         // Initialization code
+        
+        mainCellView.backgroundColor = RHColor.WhiteBackground
```
levibostian, line 26: I prefer to name colors based off of their purpose and not by the color they are. So, instead of `WhiteBackground`, instead use `HabitTableViewCellBackgroundColor`. 

I prefer this because what if one day you decide to change the color of `WhiteBackground` to a yellow color? You may have made that yellow color change because you wanted to change the background of 1 ViewController and instead of accidentally changed the color of this TableViewCell.

Remote Habits/Extension/ViewExtension.swift
```diff
@@ -1,14 +1,14 @@
...
-import SwiftUI
+import UIKit
 
-extension View {
-    func binding<Value>(_ getValue: @escaping () -> Value) -> Binding<Value> {
-        Binding.get(getValue)
+extension UIView {
+   
+    func setCornerRadius() {
+        self.layer.cornerRadius = 13
```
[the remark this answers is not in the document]
ami-aman: Done.

Remote Habits/Extension/ViewExtension.swift
ami-aman: But if you think this is not right then having an extension to "just" add corner radius should also not be appropriate. It is also just one line of code. And it is an extension (correct) but this function does not add any effect until called.

Remote Habits/View/Common/RHBaseViewController.swift
ami-aman: This was al;ready updated in the next commit. ✔️

Remote Habits/View/RHLoginViewController.swift
ami-aman: We do not need to use this in dark mode. It always has to be black in both the scenarios

Remote Habits/View/Xib/HabitDashboard/HabitTableViewCell.swift
ami-aman: haha and I do not prefer to do that because it limits the use of the property to that class/ controller whereas these colors are being used at multiple places and it feels very very odd using `HabitTableViewCellBackgroundColor` for instance in `DashboardViewController` or any other file.

Remote Habits/Extension/ViewExtension.swift
levibostian: I find value in extension functions that call 1 line of code such as only adding a corner radius. I find value in extension functions on generic classes such as `UIView`. 

I am not a fan of this code because it's an extension on the generic `UIView` class that says "call this function is you want to customize the View". It's opinionated towards all `UIView`s in it's naming and function body. 

If this extension existed in `UIHabitTableViewCell`, someless less generic, I might see more value in this specific extension with this function name. If this extension was renamed to something else that was more descriptive and less opinionated to the generic `UIView` class, I might see more value in it. 

Some suggestions off the top of my head: using a custom `UIView` that has this radius and background color set on it? Not using an extension function for this use case?

Remote Habits/View/Xib/HabitDashboard/HabitTableViewCell.swift
levibostian: > very very odd using HabitTableViewCellBackgroundColor for instance in DashboardViewController or any other file.

Yes, I agree that example would be odd and not what I was implying. I can see now that I explained it poorly re-reading my comment. 

Looking over `RHColor` again, I am ok with the current implementation as it is today because it is implying it's function already (using `WhiteBackground` instead of just saying `White`). 

What I mean by "name colors based off purpose and not color they are" I am talking about the pattern of [ semantic color[1](https://nshipster.com/dark-mode/#use-semantic-colors)[2](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/color/). j

With that pattern in mind, instead of using `WhiteBackground` which is saying the color is white (when in dark mode it's not), maybe `PrimaryBackground` would be a better name as it's saying the the primary background color of the app is `X`. If we ever decide to change the background color in the future to red, we don't need to rename the variable of `PrimaryBackground`. It stays the same. If we used `WhiteBackground` and changed the color to red, we would need to rename the variable or, create a new variable and change all of the references in the code.

ami-aman: Gotcha ! Here I agree with you !

ami-aman: I can note this down for future commits but not doing this right now because it doesn;t look like a showstopper to me. What do you say ?

Remote Habits/Extension/ViewExtension.swift
ami-aman: I would rather prefer removing it from extension and use it directly as a code in their respective classes. Gotcha your point.

Remote Habits/View/Xib/HabitDashboard/HabitTableViewCell.swift
levibostian: In my opinion, if a refactor is (1) not very time consuming and (2) is relevant to a PR - I say it's best to perform the refactor in the PR rather then introducing technical debt into the project at merge. 

Especially because this PR is heavy on color because the PR is adding dark mode, I find the refactor of renaming colors to being relevant to the PR. How long it takes is up for you to determine. 

Because you know the time commitment more then I do, I'll let you decide if you introduce the refactor now in this PR before merge or create a ticket and perform the changes later.

levibostian (approved):

PR #51, opened 2022-01-11
refactor: removed prefix RH from file names

ami-aman:
**This PR Contains :**

- Few file names in remote habits app started with prefix RH which was not aligned with the former remote habits app, hence removed prefix RH from file name.
> **Example :**
> RHLoginViewController.swift is now LoginViewController.swift

- Using enum for habit row type instead of integers [PR suggestion here](https://github.com/customerio/RemoteHabits-iOS/pull/42#discussion_r777708342)


levibostian (approved):
Renamed the title of the PR from `feat: ` to `refactor: ` since this PR isn't adding a feature but is is more of a refactoring change that doesn't change the behavior of the app.

PR #55, opened 2022-01-12
ci: modify names of test groups

levibostian:
The CI server makes builds of this app and sends to internal *groups* of testers so we can test the app/SDK. 

We are simplifying the testing groups that we use from 3 groups to 2. This PR simply changes the CI server to deploy to these new groups.


ami-aman (approved):
👍

PR #61, opened 2022-03-02
feat: configure cio sdk in app

ami-aman:
[Make it possible to connect the Remote Habits apps to a Gitpod/devbox environment](https://github.com/customerio/issues/issues/6569)

Remote Habits app is now smarter! You can now configure Customer.io SDK using Remote Habits app. Properties such as :
- Tracking URL
- Background queue delay (Number)
- Background queue minimum number of tasks (Number)
- Log level (Dropdown - debug, info, error)
- Auto track screen views (Switch true/false)
- Auto track device attributes (Switch true/false)
- Custom device attributes (JSON String such as
> "{\"foo\":\"bar\",\"appname\":\"myApp\",\"foo1\":\"bar1\",\"foo2\":\"bar2\"}"

Currently, only few properties are enabled for the reason that the current RH app has CIO SDK version 1.0.0 and all the properties are not available in this version of SDK.

Here is a look of how you can customise the SDK. The settings icon on top right of the login page can be used to configure the SDK
Simulator Screen Shot - iPhone 13 Pro Max - 2022-03-02 at 14 51 46  Simulator Screen Shot - iPhone 13 Pro Max - 2022-03-02 at 15 18 25
Simulator Screen Shot - iPhone 13 Pro Max - 2022-03-16 at 17 34 59 Simulator Screen Shot - iPhone 13 Pro Max - 2022-03-16 at 17 08 37


Remote Habits/View/ConfigureCioSdkViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+             if let data = object.data(using: .utf8) {
+             do {
+             if let dict = try JSONSerialization.jsonObject(with: data, options: []) as? [String: String] {
+             CustomerIO.shared.deviceAttributes = dict
+             }
+             } catch {
+             print(error.localizedDescription)
+             }
+             }
+             }
```
levibostian, line 109: From reading this code, is `customDeviceAttributesText` a textfield that allows you to type in custom attributes text like `["foo": "bar"]`? Is this a feature that is needed in the RH app?

Remote Habits/View/ConfigureCioSdkViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+    func findLogLevel() -> CioLogLevel {
+        switch logTypeButton.title(for: .normal) {
+        case "debug":
+            return .debug
+        case "info":
+            return .info
+        case "error":
+            return .error
+        default:
+            return .info
```
levibostian, line 85: 1. Is there a reason why we need to have the customer select a log level besides `.debug`?
2. If the answer is yes, I think the default should be `.debug` since RH is primarily used by QA and we need QA to have .debug logging on at all times to capture the most helpful logs for the mobile team.

Remote Habits/View/ConfigureCioSdkViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+             }
+             }
+             */
+
+            $0.logLevel = findLogLevel()
+            $0.autoTrackScreenViews = screenViewsSwitch.isOn
+            $0.backgroundQueueSecondsDelay = !bgQueueDelayText
+                .trimTextWithWhiteSpaces ? Double(bgQueueDelayText.text!)! : 30
+            $0.backgroundQueueMinNumberOfTasks = !bgQueueMinTasksText
+                .trimTextWithWhiteSpaces ? Int(bgQueueMinTasksText.text!)! : 10
```
levibostian, line 117: I am still feeling uneasy about changing the default background queue configuration for QA testing. 

See [this issue](https://github.com/customerio/issues/issues/6637) to see why. Most important part of that issue is: 

> Note: One solution is to change the SDK background queue configuration in the app to make the queue delay for maybe 5 seconds. The problem with this is that QA testing is meant to test the app/SDK as a customer would. The QA environment and the customer environment is now different which could lead to false positive tests and may not find some bugs.

levibostian:
I added a `blocked` label to this PR. We will merge this after the next release of the SDK which will allow modifying the tracking API URL for the SDK to send HTTP requests to.

Remote Habits.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -19,6 +19,24 @@
...
+          "branch": null,
+          "revision": "01afed4a137ac000031361a8d31dcae19e48cb42",
+          "version": "1.0.0"
+        }
+      },
+      {
+        "package": "DropDown",
+        "repositoryURL": "https://github.com/AssistoLab/DropDown.git",
+        "state": {
+          "branch": "master",
```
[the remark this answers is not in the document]
ami-aman: I do agree with your point, Levi but the tags listed in [here](https://github.com/AssistoLab/DropDown/releases) do not provide SPM support. As per AssistoLab's [pull request history](https://github.com/AssistoLab/DropDown/pulls?q=is%3Apr+is%3Aclosed), support for SPM was added in 2020 and the the last tag available is from 2019. Hence using master branch.

Remote Habits.xcodeproj/xcshareddata/xcschemes/Remote Habits.xcscheme
```diff
@@ -28,6 +28,46 @@
...
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "TrackingTests"
+               BuildableName = "TrackingTests"
+               BlueprintName = "TrackingTests"
+               ReferencedContainer = "container:../../CIO SDK/customerio-ios">
+            </BuildableReference>
+         </TestableReference>
```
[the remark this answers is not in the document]
ami-aman: Oh Oh ! I made this change but missed to discard it before pushing ! Reversing it !

Remote Habits/Extension/ButtonExtension.swift
```diff
@@ -0,0 +1,24 @@
+import UIKit
+extension UIButton {
+    func addLightGrayBorder() {
```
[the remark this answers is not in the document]
ami-aman: Great suggestion. Changes made !

Remote Habits/Extension/ButtonExtension.swift
```diff
@@ -0,0 +1,24 @@
...
+        layer.borderColor = UIColor.lightGray.withAlphaComponent(0.5).cgColor
+        layer.cornerRadius = 5
+    }
+
+    func makeFloating() {
+        let xCoordinates = Int(UIScreen.main.bounds.width - 56)
+        let widthHeight = 40
+        frame = CGRect(x: xCoordinates, y: 50, width: widthHeight, height: widthHeight)
+        setImage(UIImage(named: "settings")!, for: .normal)
+        self.layer.cornerRadius = CGFloat(widthHeight / 2)
```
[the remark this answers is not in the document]
ami-aman: The name of the function should be changed. My earlier attempt was to make this button completely floating like user could drag, move, drop it anywhere on screen but then it would have not added much value so omitted the thoughts. I am changing the name of the function to describe what it does `customiseFloating`. But since it is still on the top of the screen and appears like a floating button so would not change much except the name of function

Remote Habits/View/ConfigureCioSdkViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    // MARK: - Actions
+
+    @IBAction func updateConfigButtonTapped(_ sender: UIButton) {
+        CustomerIO.config {
+            // MARK: - Enable with latest version of SDK
+
+            /*
+             $0.trackingApiUrl = !trackingApiUrlText.trimTextWithWhiteSpaces ? trackingApiUrlText.text! : ""
```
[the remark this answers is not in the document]
ami-aman: I thought the same but then didn't find much use of adding this in the SDK! If you know somewhere we are doing a similar functionality then I would prefer moving this into SDK rather than having it here.

Remote Habits/View/ConfigureCioSdkViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+        default:
+            return .info
+        }
+    }
+
+    // MARK: - Actions
+
+    @IBAction func updateConfigButtonTapped(_ sender: UIButton) {
+        CustomerIO.config {
+            // MARK: - Enable with latest version of SDK
```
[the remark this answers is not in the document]
ami-aman: This has been discussed in March 14th sync up call. We can go ahead and merge this PR and later enable the properties as we add new SDK version.

Remote Habits/View/ConfigureCioSdkViewController.swift
ami-aman: This was not a part of the ticket but I have added so that QA team can rely less on hard coded data and play around themselves and test the SDK with the help of RH. This might sound odd at the moment but I strongly believe this will be helpful.

Remote Habits/View/ConfigureCioSdkViewController.swift
ami-aman: If you strongly believe that making changes in BG queue configurations can cause other problems then we can let go this property customisation.

Remote Habits/View/ConfigureCioSdkViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+        logTypeDropdown.anchorView = logTypeButton
+        logTypeDropdown.dataSource = ["debug", "info", "error"]
+        logTypeDropdown.selectionAction = { [unowned self] (_: Int, item: String) in
+            self.logTypeButton.setTitle(item, for: .normal)
+        }
+    }
+
+    func configureUpdateConfigButton() {
+        updateConfigButton.titleLabel?.font = Font.SFProTextSemiBoldMedium
+        updateConfigButton.setTitleColor(UIColor.white, for: .normal)
```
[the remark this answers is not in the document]
ami-aman: Added support for dark mode.

Remote Habits.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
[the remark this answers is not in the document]
levibostian: Ah, you're right. 

How about using a git commit instead of a branch? I would like to avoid using a branch. Git tag is best, but if we can't use that a git commit should work, too. Looks like the latest commit on `master` is [2ab6f6ce19f0117d1a76ea043ef8f57722c65d16](https://github.com/AssistoLab/DropDown/commit/2ab6f6ce19f0117d1a76ea043ef8f57722c65d16)

Remote Habits/View/ConfigureCioSdkViewController.swift
levibostian: For QA testing, good use case. 

Looking at a screenshot of the app, the hint for the UITextField is "Custom device attributes". It is difficult to understand what is valid to type into this text field in the app. One idea is to edit the hint for the UITextField to give an example of what to type in. 

```
hint before: "Custom device attributes"
hint after: "Custom device attributes. ["foo": "bar", "number": 2]"
```

Remote Habits/View/ConfigureCioSdkViewController.swift
levibostian: > this feature will help them test other log levels as well.
Oh, so you are imagining that the QA team would change the log level only to make sure that changing the log levels of the SDK works? 

That's a good idea. When the QA team finds a bug, I really hope that `.debug` is selected so we get the full logs to help us debug. As long as `.debug` is the default, then I am ok with being able to change the log level of the SDK. 

Maybe a code comment would be helpful explaining that (1) we added the ability to change log level only for QA testing but (2) it's important for `.debug` to be the default for debugging purposes?

Remote Habits/AppDelegate.swift
```diff
@@ -25,6 +25,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         // Step 3: Register for push notifications
         UIApplication.shared.registerForRemoteNotifications()
+        
+        CustomerIO.config {
+            $0.logLevel = .debug
+        }
```
levibostian, line 31: 1. I think that it would be more helpful if this code was inserted after `CustomerIO.initialize()` so that customers viewing this sample code would see an example of how to customize the SDK. 

```suggestion
  // Step 1: Initialise CIO SDK
  CustomerIO.initialize(siteId: workspaceId, apiKey: apiKey, region: Region.US)
  // Optionally, configure the SDK if you need to. 
   CustomerIO.config {
      $0.logLevel = .debug
   }
   // 2......
```

2. RH used to have auto screen view tracking enabled. Want to add that back to the configuration?

Remote Habits/Extension/ButtonExtension.swift
```diff
@@ -1,12 +1,12 @@
 import UIKit
 extension UIButton {
-    func addLightGrayBorder() {
+    func addColoredBorder(color : UIColor) {
         layer.borderWidth = 0.5
-        layer.borderColor = UIColor.lightGray.withAlphaComponent(0.5).cgColor
+        layer.borderColor = color.cgColor
         layer.cornerRadius = 5
```
levibostian, line 6: Adding a default corner radius is irrelevant to adding a colored border. I would move this out of this function and have each `UIButton` instance call `.setCornerRadius(5)` themselves.

levibostian (approved):
After resolving the PR conversations, I am good with having this PR merged 👍

Remote Habits.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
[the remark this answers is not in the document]
ami-aman: ✅

Remote Habits/View/ConfigureCioSdkViewController.swift
ami-aman: I will be creating a document for this particular screen and linking it to our existing "How to use RH app" notion page. That will explain everything about the expected values and how to play around with fields. 

Also, for the sake of simplicity, I have re-arranged the fields on the screen. All disabled (or yet to be introduced) properties are at the bottom. As and when we integrate newer version of SDK, I will update the screen ! It would be done in a jiffy!

Remote Habits/View/ConfigureCioSdkViewController.swift
ami-aman: I am planning to take up code documentation/commenting in Spring cleaning. Will create a task soon for that !

Remote Habits/View/ConfigureCioSdkViewController.swift
ami-aman: Okay. I totally get you here.
I would mention the same. in the document I would be creating for using these properties. How about letting it go right now and if we get issues because of customisation then we can remove this field. What say ?

Remote Habits/AppDelegate.swift
ami-aman: Here you go ✅

Remote Habits/Extension/ButtonExtension.swift
ami-aman: ✅

PR #65, opened 2022-03-23
fix: enabled configurable properties and bug fix in remote habits

ami-aman:
Made some improvements in the remote habits 

**Bug Fix on Login Screen**
This went unnoticed for long but fixed now before QA team reported 🚀 

**Bug** - Guest user always generated same user name and email Id. 
**Fix** - On clicking `Continue as a guest user` will always generate a new random email and username

Other minor improvements include - dismissing keyboard when tapped anywhere on the screen and tapping return on configure SDK screen will dismiss keyboard

**On Configure SDK screen**
- Enabled Tracking API URL field
- Set auto track device attributes and auto track screen attributes to true
- Minor changes in the UI
-  Currently, adding custom device attributes need some further improvement i.e. this configuration should be provided only when the user is logged in because if not then there will not be a profile identifier and resp function is not execute further

To provide the ability to QA, I have added a line of code in Switch Workspace where on opening the switch workspace screen , custom device attributes with some general user info (though not related to device but only for testing right now, will remove later when we add configuration property in RH) but will be random for each login. This will help us test that the custom attributes do get added and get updated randomly.

**Custom attributes added are :**
name - name of the user
current screen - which is always switch workspace
is authorised user - to find if this is a guest user or logged in user

**Successful workspace example**
Screenshot 2022-03-24 at 12 28 30 AM


Remote Habits/View/ConfigureCioSdkViewController.swift
```diff
@@ -104,25 +106,27 @@ class ConfigureCioSdkViewController: BaseViewController {
 
     @IBAction func updateConfigButtonTapped(_ sender: UIButton) {
         CustomerIO.config {
+            
+            
+            
+            $0.trackingApiUrl = !trackingApiUrlText.trimTextWithWhiteSpaces ? trackingApiUrlText.text! : ""
+            $0.autoTrackDeviceAttributes = deviceAttributesSwitch.isOn
             // MARK: - Enable with latest version of SDK
```
levibostian, line 114: Can't we enable these features now with 1.1.0 alpha?

Remote Habits/View/SwitchWorkspaceViewController.swift
```diff
@@ -26,6 +27,8 @@ class SwitchWorkspaceViewController: BaseViewController, UITextFieldDelegate {
     override func viewDidLoad() {
         super.viewDidLoad()
 
+        // Sending custom attributes to the workspace
```
levibostian, line 30: Cool!!!

Remote Habits/View/ConfigureCioSdkViewController.swift
```diff
@@ -104,25 +106,27 @@ class ConfigureCioSdkViewController: BaseViewController {
 
     @IBAction func updateConfigButtonTapped(_ sender: UIButton) {
         CustomerIO.config {
+            
+            
+
```
levibostian, line 111: I think you forgot to lint the code. These empty white spaces would probably be removed if you ran it.

levibostian (approved):

Remote Habits/View/ConfigureCioSdkViewController.swift
ami-aman: I didn't realise this earlier but this configuration should be provided only when the user is logged in because if not then there will not be a profile identifier and resp function is not execute further. I will make these changes later but soon to enable QA test this feature.

Remote Habits/View/ConfigureCioSdkViewController.swift
ami-aman: Oops! My bad ! Doing it right away before merging

Remote Habits/View/ConfigureCioSdkViewController.swift
levibostian: This line of code (the comment `// MARK: - Enable with latest version of SDK`), I believe should result in one of these action items. 

1. Enable the uncommented code below the comment and delete the comment. 
2. If we can't enable this commented code out now, create a ticket that explains when the change can be made and edit the comment in the code also explaining when the code can be enabled. If I was reading this comment as it is right now, I am not sure when I can enable this code.

ami-aman: I think I should reframe the comment and going by your suggestion - creating a ticket to tackle this in future

ami-aman: https://github.com/customerio/issues/issues/7041

PR #71, opened 2022-07-28
fix: switch workspace not working as expected 

ami-aman:
closes: https://github.com/customerio/issues/issues/7525

In Remote Habits, on entering new site id & api key, the credentials were validated but not updated correctly. The user had to force kill the app & relaunch it to see the effective change. 

This fix provides a solution to switch workspace, validate & update the credentials without the requirement of relaunching the app which was a workaround as mentioned in the ticket above

With this PR, the settings button to update the SDK configuration is now available on the Dashboard screen which was previously on the Login Screen.

### DIFF
--- a/Remote Habits/View/SwitchWorkspaceViewController.swift
+++ b/Remote Habits/View/SwitchWorkspaceViewController.swift
@@ -175,5 +175,7 @@ class SwitchWorkspaceViewController: BaseViewController, UITextFieldDelegate {
         Env.customerIOSiteId = siteId
         userManager.apiKey = apiKey
         Env.customerIOApiKey = apiKey
+        
+        CustomerIO.initialize(siteId: siteId, apiKey: apiKey, region: Region.US)
     }
 }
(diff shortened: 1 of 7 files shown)


Remote Habits/View/SwitchWorkspaceViewController.swift
```diff
@@ -175,5 +175,7 @@ class SwitchWorkspaceViewController: BaseViewController, UITextFieldDelegate {
         Env.customerIOSiteId = siteId
         userManager.apiKey = apiKey
         Env.customerIOApiKey = apiKey
+        
+        CustomerIO.initialize(siteId: siteId, apiKey: apiKey, region: Region.US)
```
jatinn, line 179: do you need to configure with the correct track url or does calling initialize not change the url?

jatinn (approved):
also think the following needs to be changed to use the specified track url if set otherwise validate creds step fails

https://github.com/customerio/RemoteHabits-iOS/blob/5c46fafb7437ca794c5cf13354416f01fc87f01e/Remote%20Habits/Repository/ProfileRepository.swift#L69

Breadcorn (approved):

levibostian:
Fixed in another PR https://github.com/customerio/RemoteHabits-iOS/pull/78

PR #72, opened 2022-08-08
feat: in-app in remote habits

ami-aman:
This is a feature PR to add in-app functionality in Remote Habits. Env file also updated with respective org id.

Simulator Screen Shot - iPhone 13 - 2022-08-08 at 18 12 58

### DIFF
--- a/Remote Habits/AppDelegate.swift
+++ b/Remote Habits/AppDelegate.swift
@@ -3,6 +3,7 @@ import CioTracking
 import CoreData
 import UIKit
 import UserNotifications
+import CioMessagingInApp
 
 @main
 class AppDelegate: UIResponder, UIApplicationDelegate {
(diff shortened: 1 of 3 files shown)


Remote Habits/AppDelegate.swift
```diff
@@ -29,6 +30,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         // Step 3: Register for push notifications
         UIApplication.shared.registerForRemoteNotifications()
+        
+        // In-app
+        MessagingInApp.shared.initialize(organizationId: Env.customerIOInAppOrganizationId)
```
jatinn, line 35: since reading from env you may also want to update the Env class example

https://github.com/customerio/RemoteHabits-iOS/blob/main/Remote%20Habits/Env.swift.example#L3

jatinn (approved):
lgtm future update might be to make this configurable from within the app

Remote Habits/AppDelegate.swift
ami-aman: Right. Thanks for pointing out. Updating Env.swift.example in [this](https://github.com/customerio/RemoteHabits-iOS/pull/72/commits/793551eb8cf1e71eb09777e70bb6126f52ddd2ce) commit.

PR #74, opened 2022-09-14
fix: app crash and workaround for image in rich push

ami-aman:
closes https://github.com/customerio/issues/issues/8112

This PR also contains workaround fix for RH not showing image in rich push notification.

### DIFF
--- a/Notification Service/NotificationService.swift
+++ b/Notification Service/NotificationService.swift
@@ -1,6 +1,6 @@
 import CioMessagingPushAPN
-import UserNotifications
 import CioTracking
+import UserNotifications
 
 class NotificationService: UNNotificationServiceExtension {
     private let cioMessagingPush: MessagingPush = DI.shared.messagingPush
--- a/Notification Service/NotificationService.swift
+++ b/Notification Service/NotificationService.swift
@@ -9,10 +9,13 @@ class NotificationService: UNNotificationServiceExtension {
         _ request: UNNotificationRequest,
         withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void
     ) {
+        // Adding temporary workaround for the issue https://github.com/customerio/customerio-ios/issues/159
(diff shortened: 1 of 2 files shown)


Remote Habits/Manager/DataManager.swift
```diff
@@ -76,6 +76,7 @@ class HabitDataManager: CoreDataManager {
                     objectUpdate.setValue(endTimeValue.toDate(withFormat: .time12Hour), forKey: "endTime")
                 }
             }
+            managedContext.mergePolicy = NSOverwriteMergePolicy
```
jatinn, line 79: maybe a comment as to why this fixes the issues

Notification Service/NotificationService.swift
```diff
@@ -11,6 +12,7 @@ class NotificationService: UNNotificationServiceExtension {
         // If you use more service then Customer.io for sending rich push messages,
         // you can check if the SDK handled the rich push for you. If it did not, you
         // know that the push was *not* sent by Customer.io and you can try another way.
+        CustomerIO.initialize(siteId: Env.customerIOSiteId, apiKey: Env.customerIOApiKey, region: Region.US)
```
jatinn, line 15: probably fine for now but we should likely be saving the region into env as well

jatinn (approved):

Remote Habits/Manager/DataManager.swift
ami-aman: ✅

Notification Service/NotificationService.swift
ami-aman: Yes, this is a temporary workaround and will be removing it once we have the changes up in the SDK. Probably I should add a comment here.

ami-aman: Added a comment mentioning the same.

PR #78, opened 2022-11-07
feat: SDK v2

levibostian:
Updates app to SDK v2
Closes: https://github.com/customerio/issues/issues/7525


Shahroz16:
I believe this is due to the fact we are making our SDK singleton, but just curious do we need to update SDK as `V3`? since we already marked those methods as `deprecated`? and we can even hack it around that every time a user tries to "create a separate instance using the deprecated method" we internally redirect the call to same `singleton` method?  We could mention that in the method as well. 

Don't have a strong opinion, just want to make sure we don't end up being on a very large SDK version or give users the impression that we make breaking changes very often.

levibostian:
Yes, [the changes I made currently introduces breaking changes and releasing a major version bump for this change](https://github.com/customerio/customerio-ios/pull/209).

matt-frizzell:
👍To Levi's comment on the version change. As we are making breaking changes, the removal of deprecated functions, developers must know that if they take this update their code could break if they are still relying on those previous items. 

In terms of perception, I do agree that we wouldn't want to misuse major versions and give the perception that we keep changing the underlying code in major ways, but the case we are talking about is appropriate. What IS good to try to do is pair something like the underlying singleton refactor with some enhancements as well. The work that is happening in customerio/issues/#8359 might be that enhancement to pair with the refactor.

Remote Habits/View/ConfigureCioSdkViewController.swift
```diff
@@ -105,9 +105,10 @@ class ConfigureCioSdkViewController: BaseViewController {
     // MARK: - Actions
 
     @IBAction func updateConfigButtonTapped(_ sender: UIButton) {
-        CustomerIO.config {
```
ami-aman, line 108: Is this also deprecated or you chose to implement it using appdelegate ? 🤔

levibostian: Also deprecated. You must now configure the SDK during the `initialize()` function instead of doing it at a later time. 

I am not remembering why we decided to make this change. I think we decided to do it before I left for parental leave so I am having a hard time remembering this change.

levibostian: Adding this for documentation reasons....

Outside of github you reminded me of why we are doing this change: for setting user-agent and for compatability with Android SDK for our RN SDK.

ami-aman (approved):

levibostian:
PR not yet ready to merge. There are bugs in v2 of the SDK that need to be deployed before this app is even able to build. 

At this time, I am trying to get the build to fail on the CI server. Since there is a known bug in 2.0.0-alpha.1 of the SDK, this app should not build successfully. I am trying to make sure the app does fail so we know we have a good test app moving forward.

PR #79, opened 2022-11-09
feat: add in-app event listener

levibostian:
- feat: add in-app event listener
- update from message opened to shown
- update lint to match SDK


lj-cio:
Update to fire an event when a listener is triggered.  Include the listener name as part of the event data so we can test all the different listeners.

PR #80, opened 2022-11-09
fix: fix in-app bug via gist 2.2.1

levibostian:


levibostian:
Bug has been fixed in SDK. Will be fixed in this app when we update the SDK version in it.

PR #82, opened 2023-01-09
ci: changes Slack deployment message channel

matt-frizzell:
Changes the Slack channel for deployment messages to the new mobile-deployments channel

See customerio/issues#9014


ami-aman (approved):
👍

levibostian (approved):

levibostian:
I don't see much value in having slack deployment messages for RH app. Also because we are deprecating the app, I think it's better for us to disable slack messages instead of trying to modify them. 

Closing. Re-open if someone believes this still adds value.

PR #88, opened 2023-06-06
ci: release stable app builds on push to main branch

levibostian:
Our TS team still uses RH. Therefore, it's important that this repo continues to publish stable builds. 

This PR [merges in the new iOS stable builds changes](https://github.com/customerio/customerio-ios/pull/332) into RH.


fastlane/Fastfile
```diff
@@ -1,6 +1,5 @@
-
 import_from_git(
   url: "https://github.com/customerio/customerio-ios.git",
-  branch: "main", # The branch to checkout on the repository.
+  branch: "levi/stable-builds-on-main", # The branch to checkout on the repository.
```
levibostian, line 3: Change to `main` [after this PR is merged](https://github.com/customerio/customerio-ios/pull/332)

.github/workflows/deploy-app.yml
```diff
@@ -1,52 +1,44 @@
...
+    - uses: actions/checkout@v3
+
+    - name: Set up Xcode to version we determine 
+      uses: maxim-lobanov/setup-xcode@v1
+      with:
+        xcode-version: 14
+
+    - name: Install CLI tools used in CI script 
+      run: |
+        brew install xcbeautify # used by fastlane for output
```
Shahroz16, line 24: can we somehow move this dependency in the fastlane setup? rather than github action workflow? because we have to duplicate it in all workflows.

Shahroz16 (approved):

.github/workflows/deploy-app.yml
levibostian: I agree that duplication in all of the workflows exists and is not ideal. 

Installing this tool is not the only place that we are duplicating work. We are also duplicating installing Android SDK and Xcode. 

Therefore, I like the solution of creating re-usable GitHub Action for our mobile team that does this installing. 

```yml
jobs:
  build-android-app:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - uses: customerio/mobile-ci/ios/build-setup@main
```

[I have that work as part of this ticket](https://github.com/customerio/issues/issues/9019#issuecomment-1376268269).

